PR #10, opened 2024-05-16
Implementation of swing filter and swing mid-range 

cmcuza:
In this PR, the algorithm swing-filter [1] is implemented. Moreover, a version of it called swing-mid-range has also been implemented. The main difference is that swing-filter minimizes the square error between the approximation line and the uncompressed values. Instead, swing-mid-range only finds the line that crosses the middle point of the upper and lower bounds lines. This PR contains several tests for both algorithms. The PR contains code from Poor's Man Compression to be able to run properly.  

[1]Hazem Elmeleegy, Ahmed K. Elmagarmid, Emmanuel Cecchet, Walid G. Aref, and Willy Zwaenepoel. Online piece-wise linear approximation of numerical streams with precision guarantees. Proc. VLDB Endow. 2, 1, 2009. https://doi.org/10.14778/1687627.1687645.

### DIFF
--- a/.github/workflows/zig-build-lint-and-test-on-pr-and-push.yml
+++ b/.github/workflows/zig-build-lint-and-test-on-pr-and-push.yml
@@ -26,6 +26,7 @@ jobs:
     name: Zig Build and Test
 
     runs-on: ${{ matrix.operating-system }}
+    timeout-minutes: 10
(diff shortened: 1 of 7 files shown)


cmcuza:
PR closed because it was superseded by another PR

PR #15, opened 2024-06-02
Restructure the segment structure

cmcuza:
This PR restructures the Segment structure to use the Point structure for its attributes start_point and end_point. All functions using the Segment structure have been updated accordingly as well as documentation.

### DIFF
--- a/src/functional/swing_slide_filter.zig
+++ b/src/functional/swing_slide_filter.zig
@@ -278,8 +278,8 @@ fn appendValue(comptime T: type, value: T, compressed_values: *std.ArrayList(u8)
     }
 }
 
-/// Computes the intercept coefficient of a linear function that passes through `point` and
-/// with the given slope coefficient.
+/// Computes the intercept coefficient of a linear function that passes through `point`
+/// with the given `slope` coefficient.
 fn computeInterceptCoefficient(slope: f80, point: Point) f80 {
     return point.value - slope * usizeToF80(point.time);
 }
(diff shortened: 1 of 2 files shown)


src/functional/swing_slide_filter.zig
```diff
@@ -290,10 +278,10 @@ fn appendValue(comptime T: type, value: T, compressed_values: *std.ArrayList(u8)
     }
 }
 
-/// Computes the intercept coefficient of a linear function that passes through the point
-/// (time, value) and with the given slope coefficient.
-fn computeInterceptCoefficient(slope: f80, time: usize, value: f64) f80 {
-    return value - slope * usizeToF80(time);
+/// Computes the intercept coefficient of a linear function that passes through `point` and
+/// with the given slope coefficient.
```
skejserjensen, line 282: Very minor, but either with or and. Also slope lacks backticks.

src/tersets.zig
```diff
@@ -39,6 +39,9 @@ pub const Method = enum {
 /// A point represented by `time` and `value`.
 pub const Point = struct { time: usize, value: f64 };
 
+/// Segment represented by the `start_point` and `end_point`.
```
skejserjensen, line 42: Maybe state what a segment is before stating what it is represented by. I do not think it is as self explanatory as point.

skejserjensen (changes requested):

src/functional/swing_slide_filter.zig
cmcuza: Fixed.

src/tersets.zig
cmcuza: This is tricky. Finding a proper definition of a Segment without creating confusion is not straightforward. 
Turns out that the proper definition is:  Part of a line that is bounded by two distinct endpoints.
However, that inserts the concept line, which can be problematic. I propose the following:
/// This structure models a straight line segment from `start_point` to `end_point`.

skejserjensen (approved):

PR #38, opened 2024-06-10
Adapt Point structure to a continous time axis

cmcuza:
This PR is related to #13 as "Slide Filter" needs a continuous time axis and currently, we only support discrete time (x-axis). There are two ways to address this problem. 1) Change the Point structure and the implementation of "Swing Filter" to properly utilize the new continuous time axis. 2) Create a new Point structure called ContinousPoint and make "Slide Filter" utilize only this ContinousPoint. I decided the first one is better although it creates changes all around in the Swing Filter's compress and decompress functions. After testing, everything is working well and the new Point is working as expected.

### DIFF
--- a/src/functional/swing_slide_filter.zig
+++ b/src/functional/swing_slide_filter.zig
@@ -70,20 +70,24 @@ pub fn compressSwing(
     var current_timestamp: usize = 2;
     while (current_timestamp < uncompressed_values.len) : (current_timestamp += 1) {
         // Evaluate the upper and lower bound linear functions at current timestamp.
-        const upper_limit = evaluateLinearFunctionAtTime(upper_bound_linear_function, current_timestamp);
-        const lower_limit = evaluateLinearFunctionAtTime(lower_bound_linear_function, current_timestamp);
+        const upper_limit = evaluateLinearFunctionAtTime(
+            upper_bound_linear_function,
+            @floatFromInt(current_timestamp),
+        );
+        const lower_limit = evaluateLinearFunctionAtTime(
+            lower_bound_linear_function,
+            @floatFromInt(current_timestamp),
+        );
 
         if ((upper_limit < (uncompressed_values[current_timestamp] - adjusted_error_bound)) or
             (lower_limit > (uncompressed_values[current_timestamp] + adjusted_error_bound)))
         {
             // Recording mechanism (the current points is outside the limits).
             try appendValue(f64, current_segment.start_point.value, compressed_values);
-            const segment_size = current_timestamp - current_segment.start_point.time - 1;
+            const segment_size = @as(f64, @floatFromInt(current_timestamp)) - current_segment.start_point.time - 1;
             if (segment_size > 1) {
                 // Denominator of Eq. (6).
-                const sum_square: f80 = @floatFromInt(
-                    segment_size * (segment_size + 1) * (2 * segment_size + 1) / 6,
-                );
+                const sum_square: f80 = segment_size * (segment_size + 1) * (2 * segment_size + 1) / 6;
 
                 // Get optimal slope that minimizes the squared error Eq. (5).
                 const slope: f80 = @max(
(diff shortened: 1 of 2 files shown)


src/functional/swing_slide_filter.zig
```diff
@@ -70,20 +70,24 @@ pub fn compressSwing(
     var current_timestamp: usize = 2;
     while (current_timestamp < uncompressed_values.len) : (current_timestamp += 1) {
         // Evaluate the upper and lower bound linear functions at current timestamp.
-        const upper_limit = evaluateLinearFunctionAtTime(upper_bound_linear_function, current_timestamp);
-        const lower_limit = evaluateLinearFunctionAtTime(lower_bound_linear_function, current_timestamp);
+        const upper_limit = evaluateLinearFunctionAtTime(
+            upper_bound_linear_function,
+            @floatFromInt(current_timestamp),
```
skejserjensen, line 75: Can the use of `f64` for indices cause problems such a `f64` being unable to represent an `usize` value since there theoretically can be up to `usize` values to compress and `@bitSizeOf(f64) == @bitSizeOf(usize)` on some platforms, e.g., x86-64. Also if a problem occurs will it be detected and an error returned or will incorrectly compressed data be returned? If incorrectly compressed will be returned, additional checks should be added to prevent this even if it adds a significant amount of code complexity and reduce runtime performance as a compression method should never return corrupted data without making it clear that an error has occurred. If a significant amount of code complexity and runtime checks are required to make continues time safe to use in Swing, maybe revert back to using discrete time for Swing and add `ContinousPoint` for Slide.

src/functional/swing_slide_filter.zig
```diff
@@ -70,20 +70,24 @@ pub fn compressSwing(
...
+        );
 
         if ((upper_limit < (uncompressed_values[current_timestamp] - adjusted_error_bound)) or
             (lower_limit > (uncompressed_values[current_timestamp] + adjusted_error_bound)))
         {
             // Recording mechanism (the current points is outside the limits).
             try appendValue(f64, current_segment.start_point.value, compressed_values);
-            const segment_size = current_timestamp - current_segment.start_point.time - 1;
+            const segment_size = @as(f64, @floatFromInt(current_timestamp)) - current_segment.start_point.time - 1;
```
skejserjensen, line 87: If continuous time is kept in Swing, maybe extend the doc comment of the module to make it clear why. The use of an `f64` for time in the form of indices as specified in #34 and all of the `@floatFromInt()` casts looks very strange without this information.

src/functional/swing_slide_filter.zig
```diff
@@ -303,15 +303,16 @@ test "swing filter zero error bound and even size compress and decompress" {
...
+    var i: f64 = 0;
+    while (i < 10) : (i += 1) {
         const noise = rnd.random().float(f64) * 0.1 - 0.05;
         try list_values.append(evaluateLinearFunctionAtTime(linear_function, i) + noise);
     }
 
     const uncompressed_values = list_values.items;
 
     try compressSwing(uncompressed_values[0..], &compressed_values, error_bound);
+
```
skejserjensen, line 315: It is unclear why an empty line was added here.

src/tersets.zig
```diff
@@ -38,7 +38,7 @@ pub const Method = enum {
 };
 
 /// A point represented by `time` and `value`.
-pub const Point = struct { time: usize, value: f64 };
+pub const Point = struct { time: f64, value: f64 };
```
skejserjensen, line 41: Maybe make `Point` and `Segment` generic like [`Array`](https://github.com/cmcuza/TerseTS/blob/4c4dc15a68fd1146377864e9b4e12779329f6c83/src/capi.zig#L96) so the different compression methods can use either discrete or continuous time depending on their need.

src/tersets.zig
```diff
@@ -128,13 +128,17 @@ pub fn isWithinErrorBound(
     error_bound: f32,
 ) bool {
     if (uncompressed_values.len != decompressed_values.len) {
+        std.debug.print("\nHere not the same size\n", .{});
```
skejserjensen, line 131: Maybe remove the leftover debugging code.

src/tersets.zig
```diff
@@ -128,13 +128,17 @@ pub fn isWithinErrorBound(
...
         return false;
     }
 
     for (0..uncompressed_values.len) |index| {
         const uncompressed_value = uncompressed_values[index];
         const decompressed_value = decompressed_values[index];
-        if (@abs(uncompressed_value - decompressed_value) > error_bound) return false;
+        if (@abs(uncompressed_value - decompressed_value) > error_bound) {
+            std.debug.print("\n{} and {}\n", .{ uncompressed_value, decompressed_value });
```
skejserjensen, line 139: Maybe remove the leftover debugging code and make the `if` expression a one-liner again.

skejserjensen (changes requested):

src/tersets.zig
cmcuza: Ummm that is a better idea.

src/tersets.zig
cmcuza: Ups, did not see this. My bad!

src/tersets.zig
cmcuza: Ohh no, I forgot about it.

src/functional/swing_slide_filter.zig
cmcuza: Indeed, I think I will follow this comment https://github.com/cmcuza/TerseTS/pull/38#discussion_r1633848505, close this PR, and create another where Swing does not change its implementation. Moreover, since Point will be a function, I will need to make the Segments a function. This deserves its PR.

cmcuza:
Following https://github.com/cmcuza/TerseTS/pull/38#discussion_r1633848505, I close this PR. A new PR with the corresponding changes and reverting the code in Swing is needed.

Issue #30, opened 2024-06-02
Implement Swing-Filter, and Slide-Filter


cmcuza:

cmcuza:
The swing filter method is already implemented and merged in the main branch.

skejserjensen:
Consider looking into if the internals of `decompressSwing()` and `decompressSlide()` overlap enough that it can be moved to a generic function called by `decompressSwing()` and `decompressSlide()` as discussed [here](https://github.com/cmcuza/TerseTS/pull/39#discussion_r1634706299).

PR #39, opened 2024-06-11
Add generic Point and Segment structure

cmcuza:
This PR implements a generic function to create Points and Segments in TerseTS. Currently, "Swing Filter" only requires discrete time-axis points and segments, however, "Slide Filter" requires continuous time-axis points and segments. Based on the comments made by @skejserjensen in https://github.com/cmcuza/TerseTS/pull/38#issue-2344112220, such a generic function allows us to create more effective structures depending on the needs of each compression method. Thus, https://github.com/cmcuza/TerseTS/pull/38#issue-2344112220 was closed and this new PR was created.

### DIFF
--- a/src/functional/geometry/convex_hull.zig
+++ b/src/functional/geometry/convex_hull.zig
@@ -25,14 +25,15 @@ const ArrayList = std.ArrayList;
 const testing = std.testing;
 
 const tersets = @import("../../tersets.zig");
-const Point = tersets.Point;
+const Point = tersets.DiscretePoint;
 
 /// Enum for the angle's `Turn` of three consecutive points A, B, and C. Essentially, it describes
 /// whether the path from A to B to C makes a `left` turn, a `right` turn, or continues in a
 /// straight line also called collinear.
 const Turn = enum(i8) { right, left, collinear };
 
-/// Convex Hull formed by an upper and lower hull.
+/// Convex Hull formed by an upper and lower hull. The hulls are formed by the input data with
+/// discrete time axis thus, `Point` is always represented by a `DiscretePoint`.
 pub const ConvexHull = struct {
     lower_hull: ArrayList(Point),
     upper_hull: ArrayList(Point),
--- a/src/functional/geometry/convex_hull.zig
+++ b/src/functional/geometry/convex_hull.zig
@@ -25,7 +25,7 @@ const ArrayList = std.ArrayList;
 const testing = std.testing;
 
 const tersets = @import("../../tersets.zig");
-const Point = tersets.DiscretePoint;
+const DiscretePoint = tersets.DiscretePoint;
 
 /// Enum for the angle's `Turn` of three consecutive points A, B, and C. Essentially, it describes
 /// whether the path from A to B to C makes a `left` turn, a `right` turn, or continues in a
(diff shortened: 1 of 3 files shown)


src/functional/geometry/convex_hull.zig
```diff
@@ -25,14 +25,15 @@ const ArrayList = std.ArrayList;
 const testing = std.testing;
 
 const tersets = @import("../../tersets.zig");
-const Point = tersets.Point;
+const Point = tersets.DiscretePoint;
```
skejserjensen, line 28: Maybe replace with `const Point` with `const DiscretePoint` like in `swing_slide_filter.zig` to make it explicit which version is used and make the code consistent.

src/functional/geometry/convex_hull.zig
```diff
@@ -25,14 +25,15 @@ const ArrayList = std.ArrayList;
...
 
 /// Enum for the angle's `Turn` of three consecutive points A, B, and C. Essentially, it describes
 /// whether the path from A to B to C makes a `left` turn, a `right` turn, or continues in a
 /// straight line also called collinear.
 const Turn = enum(i8) { right, left, collinear };
 
-/// Convex Hull formed by an upper and lower hull.
+/// Convex Hull formed by an upper and lower hull. The hulls are formed by the input data with
+/// discrete time axis thus, `Point` is always represented by a `DiscretePoint`.
```
skejserjensen, line 36: Maybe just replace `Point` with `DiscretePoint` in the code.

src/functional/swing_slide_filter.zig
```diff
@@ -210,7 +210,7 @@ pub fn decompress(
     var first_timestamp: usize = 0;
     var index: usize = 0;
     while (index < compressed_lines_and_index.len) : (index += 3) {
-        const current_segment: Segment = .{
+        const current_segment: DiscreteSegment = .{
```
skejserjensen, line 213: Consider looking into if the internals of `decompressSwing()` and `decompressSlide()` overlap enough that it can be moved to a generic function called by `decompressSwing()` and `decompressSlide()`.

src/functional/swing_slide_filter.zig
```diff
@@ -278,9 +278,9 @@ fn appendValue(comptime T: type, value: T, compressed_values: *std.ArrayList(u8)
     }
 }
 
-/// Computes the intercept coefficient of a linear function that passes through `point`
-/// with the given `slope` coefficient.
-fn computeInterceptCoefficient(slope: f80, point: Point) f80 {
+/// Computes the intercept coefficient of a linear function that passes through the `DiscretePoint`
```
skejserjensen, line 281: Maybe remove `DiscretePoint` from the comment for consistency as we only mention that parameter names and not the type elsewhere.

src/tersets.zig
```diff
@@ -150,3 +150,17 @@ pub fn getMaxMethodIndex() usize {
 
     return max_index;
 }
+
+/// `Point` is a point represented by `time` and `value` of datatype `time_type` and `value_type`.
+fn Point(comptime time_type: type, comptime value_type: type) type {
```
skejserjensen, line 155: Unclear why the value type is generic instead of always `f64` like everywhere else in the library.

src/tersets.zig
```diff
@@ -150,3 +150,17 @@ pub fn getMaxMethodIndex() usize {
...
+    return struct { time: time_type, value: value_type };
+}
+
+/// `Segment` models a straight line segment from `start_point` to `end_point`. The `time` and
+/// `value` datatypes of the points are determined by `time_type` and `value_type`.
+fn Segment(comptime time_type: type, comptime value_type: type) type {
```
skejserjensen, line 161: Unclear why the value type is generic instead of always `f64` like everywhere else in the library.

skejserjensen (changes requested):

src/tersets.zig
cmcuza: I did it generic to provide more flexibility but it is---indeed---so far, always f64.

src/functional/geometry/convex_hull.zig
cmcuza: Fixed.

src/functional/geometry/convex_hull.zig
cmcuza: Fixed.

src/functional/swing_slide_filter.zig
cmcuza: Fixed.

src/functional/swing_slide_filter.zig
skejserjensen: Ok, I have added a comment with a link to this discussion to #30.

skejserjensen (approved):

PR #43, opened 2024-06-19
Add Sim-Piece

### LINKED ISSUE #19: Implement Sim-Piece

cmcuza:
This PR adds the Sim-Piece compression algorithm from the paper [1] and Closes #19. The implementation is based on the authors' implementation at https://github.com/xkitsios/Sim-Piece (accessed 19-06-24). Few changes were made to support
an error bound equal to zero and to improve numerical stability. This is because Sim-Piece does not support lossless compression. Setting `error_bound` to zero will cause an error due to undefined quantization (`b=floor(value/error_bound)*error_bound`). This has a ripple effect across the algorithm that demands the implementation of special instructions to handle a zero error bound. Nevertheless, the numerical instabilities inherent to floating point operations mean that decompressed values will not exactly match the original uncompressed values. To alert the user, a warning is shown if the `error_bound` equals zero. By testing the method using an error bound equal to zero I found that decompressed values are at least approximately 1e-6 the uncompressed values. 

[1] https://doi.org/10.14778/3594512.3594521

### DIFF
--- a/src/functional/sim_piece.zig
+++ b/src/functional/sim_piece.zig
@@ -1,0 +1,113 @@
+// Copyright 2024 TerseTS Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Implementation of the Sim-Piece algorithm from the paper
+//! "Xenophon Kitsios, Panagiotis Liakos, Katia Papakonstantinopoulou, and Yannis Kotidis.
+//! Sim-Piece: Highly Accurate Piecewise Linear Approximation through Similar Segment Merging.
+//! Proc. VLDB Endow. 16, 8 2023.
+//! https://doi.org/10.14778/3594512.3594521".
+
+const std = @import("std");
+const math = std.math;
+const mem = std.mem;
+const testing = std.testing;
+const ArrayList = std.ArrayList;
+
+const tersets = @import("../tersets.zig");
+const Error = tersets.Error;
+const DiscretePoint = tersets.DiscretePoint;
+
+const SimPieceContainer = struct {
+    timestamp: usize,
+    intercept: f64,
+    upper_bound_slope: f64,
+    lower_bound_slope: f64,
+};
+
+/// Compress `uncompressed_values` within `error_bound` using "Sim-Piece" and write the
+/// result to `compressed_values`. If an error occurs it is returned.
+pub fn compressSimPiece(
+    uncompressed_values: []const f64,
+    compressed_values: *ArrayList(u8),
+    allocator: mem.Allocator,
+    error_bound: f32,
+) Error!void {
+    // Adjust the error bound to avoid exceeding it during decompression.
+    const adjusted_error_bound = if (error_bound > 0)
+        error_bound - tersets.ErrorBoundMargin
+    else
+        error_bound;
+
+    var simpiece_container_list = ArrayList(SimPieceContainer).init(allocator);
+    defer simpiece_container_list.deinit();
+
+    var upper_bound_slope: f64 = math.f64_max;
+    var lower_bound_slope: f64 = math.f64_min;
+
+    // Initialize the `start_point` with the first uncompressed value.
+    var start_point: DiscretePoint = .{ .time = 0, .value = uncompressed_values[0] };
+
+    var quantized_start_value = quantize(uncompressed_values[0], adjusted_error_bound);
+
+    // First point already part of `current_segment`, next point is at index one.
+    var current_timestamp: usize = 1;
+    while (current_timestamp < uncompressed_values.len) : (current_timestamp += 1) {
+        const end_point: DiscretePoint = .{
+            .time = current_timestamp,
+            .value = uncompressed_values[current_timestamp],
+        };
+
+        const segment_size: usize = current_timestamp - start_point.time;
+        const upper_limit: f64 = upper_bound_slope * segment_size + quantized_start_value;
+        const lower_limit: f64 = lower_bound_slope * segment_size + quantized_start_value;
+
+        if (segment_size > 2) {
+            if ((upper_limit < (end_point.value - adjusted_error_bound)) or
+                ((lower_limit > (end_point.value + adjusted_error_bound))))
+            {
+                simpiece_container_list.append(.{
+                    .timestamp = start_point.time,
+                    .intercept = quantized_start_value,
+                    .upper_bound_slope = upper_bound_slope,
+                    .lower_bound_slope = lower_bound_slope,
+                });
+                start_point = end_point;
+                quantized_start_value = quantize(start_point.value, adjusted_error_bound);
+                continue;
+            }
+        }
+
+        const new_upper_bound_slope: f64 =
+            (end_point.value + adjusted_error_bound - quantized_start_value) / segment_size;
+        const new_lower_bound_slope: f64 =
+            (end_point.value - adjusted_error_bound - quantized_start_value) / segment_size;
+        if (segment_size == 2) {
+            upper_bound_slope = new_upper_bound_slope;
+            lower_bound_slope = new_lower_bound_slope;
+        } else {
+            if (end_point.value + adjusted_error_bound < upper_limit)
+                upper_bound_slope = @max(new_upper_bound_slope, lower_limit);
+            if (end_point.value - adjusted_error_bound > lower_limit)
+                lower_bound_slope = @min(new_lower_bound_slope, upper_limit);
+        }
+    }
+}
+
+// fn compressSimPiecePhaseOne()
+
+/// Quantizes the given `value` by the specified `error_bound`. This process ensures that
+/// the quantized value remains within the error bound of the original value.
+fn quantize(value: f64, error_bound: f32) f64 {
+    return @floor(value / error_bound) * error_bound;
+}
--- a/src/functional/sim_piece.zig
+++ b/src/functional/sim_piece.zig
@@ -487,7 +487,7 @@ fn appendSegmentMetadata(
     try get_result.value_ptr.*.append(metadata);
 }
 
-/// Returns a comparator function that compares SegmentMetadata by `lower_bound_slope`.
+/// Compares `metadata_one` and `metadata_two` by their lower bound slope.
 fn compareMetadataBySlope(
     _: void,
     metadata_one: SegmentMetadata,
@@ -496,7 +496,7 @@ fn compareMetadataBySlope(
     return metadata_one.lower_bound_slope < metadata_two.lower_bound_slope;
 }
 
-/// Returns a comparator function that compares SegmentMetadata by `start_time`.
+/// Compares `metadata_one` and `metadata_two` by their start time.
 fn compareMetadataByStartTime(
     _: void,
     metadata_one: SegmentMetadata,
--- a/src/functional/sim_piece.zig
+++ b/src/functional/sim_piece.zig
@@ -59,57 +59,47 @@ pub fn compressSimPiece(
     allocator: mem.Allocator,
     error_bound: f32,
 ) Error!void {
-    // Adjust the error bound to avoid exceeding it during decompression.
-    const adjusted_error_bound = if (error_bound > 0)
-        error_bound - tersets.ErrorBoundMargin
-    else
-        error_bound;
-
     var segments_metadata_map = HashMap(
         f64,
         ArrayList(SegmentMetadata),
         HashF64Context,
         std.hash_map.default_max_load_percentage,
     ).init(allocator);
-    defer segments_metadata_map.deinit();
+    defer {
+        var iterator = segments_metadata_map.iterator();
+        while (iterator.next()) |entry| {
+            entry.value_ptr.*.deinit();
+        }
+        segments_metadata_map.deinit();
+    }
 
     // Sim-Piece Phase 1: Compute the Segment Metadata Map.
     try computeSegmentsMetadataMap(
         uncompressed_values,
         &segments_metadata_map,
         allocator,
-        adjusted_error_bound,
+        error_bound,
     );
 
-    // var merged_segments_metadata = ArrayList(SegmentMetadata).init(allocator);
-    // merged_segments_metadata.deinit();
+    var merged_segments_metadata = ArrayList(SegmentMetadata).init(allocator);
+    defer merged_segments_metadata.deinit();
 
     // Sim-Piece Phase 2: Merge the segments.
-    // try computeMergedSegmentsMetadata(
-    //     segments_metadata_map,
-    //     &merged_segments_metadata,
-    //     allocator,
-    // );
+    try computeMergedSegmentsMetadata(segments_metadata_map, &merged_segments_metadata, allocator);
 
-    var reshaped_segments_metadata = AutoHashMap(usize, HashMap(
-        f64,
-        ArrayList(usize),
-        HashF64Context,
-        std.hash_map.default_max_load_percentage,
-    )).init(allocator);
-    reshaped_segments_metadata.deinit();
+    // var reshaped_segments_metadata = AutoHashMap(usize, HashMap(
+    //     f64,
+    //     ArrayList(usize),
+    //     HashF64Context,
+    //     std.hash_map.default_max_load_percentage,
+    // )).init(allocator);
+    // reshaped_segments_metadata.deinit();
 
     // Sim-Piece Phase 3: Reshape segment metadata to output.
     // reshapeSegmentsMetadata(
     //     merged_segments_metadata,
     //     &reshaped_segments_metadata,
     // );
-
-    var iterator = segments_metadata_map.iterator();
-    while (iterator.next()) |entry| {
-        entry.value_ptr.deinit();
-    }
-
     try compressed_values.append(0);
 }
 
@@ -124,15 +114,21 @@ fn computeSegmentsMetadataMap(
         std.hash_map.default_max_load_percentage,
     ),
     allocator: mem.Allocator,
-    adjusted_error_bound: f32,
+    error_bound: f32,
 ) !void {
+    // Adjust the error bound to avoid exceeding it during decompression.
+    const adjusted_error_bound = if (error_bound > 0)
+        error_bound - tersets.ErrorBoundMargin
+    else
+        error_bound;
+
     var upper_bound_slope: f64 = math.floatMax(f64);
-    var lower_bound_slope: f64 = math.floatMin(f64);
+    var lower_bound_slope: f64 = -math.floatMax(f64);
 
     // Initialize the `start_point` with the first uncompressed value.
     var start_point: DiscretePoint = .{ .time = 0, .value = uncompressed_values[0] };
 
-    var quantized_start_value = quantize(uncompressed_values[0], adjusted_error_bound);
+    var quantized_start_value = quantize(uncompressed_values[0], error_bound);
 
     // First point already part of `current_segment`, next point is at index one.
     var current_timestamp: usize = 1;
@@ -142,22 +138,18 @@ fn computeSegmentsMetadataMap(
             .value = uncompressed_values[current_timestamp],
         };
 
-        const segment_size: usize = current_timestamp - start_point.time;
-        const upper_limit: f64 = upper_bound_slope * @as(
-            f64,
-            @floatFromInt(segment_size),
-        ) + quantized_start_value;
-
-        const lower_limit: f64 = lower_bound_slope * @as(
-            f64,
-            @floatFromInt(segment_size),
-        ) + quantized_start_value;
+        const segment_size: f64 = @floatFromInt(current_timestamp - start_point.time);
+        const upper_limit: f64 = upper_bound_slope * segment_size + quantized_start_value;
+        const lower_limit: f64 = lower_bound_slope * segment_size + quantized_start_value;
 
         if ((upper_limit < (end_point.value - adjusted_error_bound)) or
             ((lower_limit > (end_point.value + adjusted_error_bound))))
         {
             // The new point is outside the upper and lower limit. Record a new segment metadata in
             // `segments_metadata_map` associated to `quantized_start_value`.
+            std.debug.print("***Recording, start point {} ", .{start_point.time});
+            std.debug.print("upper and lower bounds {:.4} {:.4} ", .{ upper_bound_slope, lower_bound_slope });
+            std.debug.print("intercept {:.4} \n", .{quantized_start_value});
             try addSegmentMetadata(segments_metadata_map, .{
                 .start_time = start_point.time,
                 .upper_bound_slope = upper_bound_slope,
@@ -165,27 +157,33 @@ fn computeSegmentsMetadataMap(
             }, quantized_start_value, allocator);
 
             start_point = end_point;
-            quantized_start_value = quantize(start_point.value, adjusted_error_bound);
+            quantized_start_value = quantize(start_point.value, error_bound);
             upper_bound_slope = math.floatMax(f64);
-            lower_bound_slope = math.floatMin(f64);
+            lower_bound_slope = -math.floatMax(f64);
         } else {
             // The new point is within the upper and lower bounds. Update the bounds' slopes.
+
             const new_upper_bound_slope: f64 =
-                (end_point.value + adjusted_error_bound - quantized_start_value) /
-                @as(f64, @floatFromInt(segment_size));
+                (end_point.value + adjusted_error_bound - quantized_start_value) / segment_size;
             const new_lower_bound_slope: f64 =
-                (end_point.value - adjusted_error_bound - quantized_start_value) /
-                @as(f64, @floatFromInt(segment_size));
+                (end_point.value - adjusted_error_bound - quantized_start_value) / segment_size;
 
             if (end_point.value + adjusted_error_bound < upper_limit)
-                upper_bound_slope = @max(new_upper_bound_slope, lower_limit);
+                upper_bound_slope = @max(new_upper_bound_slope, lower_bound_slope);
             if (end_point.value - adjusted_error_bound > lower_limit)
-                lower_bound_slope = @min(new_lower_bound_slope, upper_limit);
+                lower_bound_slope = @min(new_lower_bound_slope, upper_bound_slope);
         }
     }
 
-    const segment_size: usize = current_timestamp - start_point.time;
-    if (segment_size != 0) {
+    const segment_size = current_timestamp - start_point.time;
+    if (segment_size >= 1) {
+        if (segment_size == 1) {
+            upper_bound_slope = 0;
+            lower_bound_slope = 0;
+        }
+        std.debug.print("***Recording, start point {} ", .{start_point.time});
+        std.debug.print("upper and lower bounds {:.4} {:.4} ", .{ upper_bound_slope, lower_bound_slope });
+        std.debug.print("intercept {:.4} \n", .{quantized_start_value});
         try addSegmentMetadata(segments_metadata_map, .{
             .start_time = start_point.time,
             .upper_bound_slope = upper_bound_slope,
@@ -208,8 +206,12 @@ fn computeMergedSegmentsMetadata(
     var timestamps_array = ArrayList(usize).init(allocator);
     defer timestamps_array.deinit();
 
+    std.debug.print("\n", .{});
+    std.debug.print("Count of segments {}\n", .{segments_metadata_map.count()});
     var iterator = segments_metadata_map.iterator();
+
     while (iterator.next()) |entry| {
+        std.debug.print("B={} \n", .{entry.key_ptr.*});
         const metadata_array = entry.value_ptr;
         mem.sort(
             SegmentMetadata,
@@ -239,7 +241,7 @@ fn computeMergedSegmentsMetadata(
             } else {
                 // A new merged segment metadata needs to be created.
                 for (timestamps_array.items) |timestamp| {
-                    try merged_segments_metadata.append(.{
+                    try merged_segments_metadata.append(SegmentMetadata{
                         .start_time = timestamp,
                         .lower_bound_slope = merge_metadata.lower_bound_slope,
                         .upper_bound_slope = merge_metadata.upper_bound_slope,
@@ -248,6 +250,7 @@ fn computeMergedSegmentsMetadata(
             }
         }
     }
+    std.debug.print("Count of segments {}\n", .{segments_metadata_map.count()});
 }
 
 /// Sim-Piece Phase 3. Reshape the input segment metadata array.
@@ -270,12 +273,11 @@ fn addSegmentMetadata(
     quantize_value: f64,
     allocator: mem.Allocator,
 ) !void {
-    const result = try metadata_map.getOrPut(quantize_value);
-    if (!result.found_existing) {
-        result.value_ptr.* = ArrayList(SegmentMetadata).init(allocator);
+    const get_result = try metadata_map.getOrPut(quantize_value);
+    if (!get_result.found_existing) {
+        get_result.value_ptr.* = ArrayList(SegmentMetadata).init(allocator);
     }
-    var metadata_array = result.value_ptr.*;
-    try metadata_array.append(metadata);
+    try get_result.value_ptr.*.append(metadata);
 }
 
 fn compareMetadata(comptime T: type) fn (void, T, T) bool {
@@ -392,42 +394,29 @@ test "hashmap can map f64 to segment metadata array list" {
     }
 }
 
-// test "sim-piece can compress and decompress" {
-//     const allocator = testing.allocator;
+test "sim-piece can compress and decompress" {
+    const allocator = testing.allocator;
 
-//     var list_values = ArrayList(f64).init(allocator);
-//     defer list_values.deinit();
-//     var compressed_values = ArrayList(u8).init(allocator);
-//     defer compressed_values.deinit();
-//     // const error_bound: f32 = 2.0;
-//     var segments_metadata_map = HashMap(
-//         f64,
-//         f64,
-//         Hash64Context,
-//         std.hash_map.default_max_load_percentage,
-//     ).init(allocator);
-//     defer segments_metadata_map.deinit();
+    var list_values = ArrayList(f64).init(allocator);
+    defer list_values.deinit();
+    var compressed_values = ArrayList(u8).init(allocator);
+    defer compressed_values.deinit();
+    const error_bound: f32 = 2;
 
-//     var rnd = std.rand.DefaultPrng.init(0);
-//     std.debug.print("\n", .{});
-//     for (0..20) |_| {
-//         try list_values.append(@floor(rnd.random().float(f64) * 1000) / 100);
-//         try segments_metadata_map.put(list_values.getLast(), list_values.getLast());
-//         std.debug.print("{} ", .{list_values.getLast()});
-//     }
-//     std.debug.print("\n", .{});
+    var rnd = std.rand.DefaultPrng.init(2);
+    std.debug.print("\n", .{});
+    for (0..10) |_| {
+        try list_values.append(@floor(rnd.random().float(f64) * 100) / 10);
+        std.debug.print("{} ", .{list_values.getLast()});
+    }
+    std.debug.print("\n", .{});
 
-//     const res = segments_metadata_map.get(list_values.items[3]);
+    const uncompressed_values = list_values.items;
 
-//     std.debug.print("Element test {}\n", .{res.?});
-//     std.debug.print("\n", .{});
-
-//     // const uncompressed_values = list_values.items;
-
-// try compressSimPiece(
-//     uncompressed_values[0..],
-//     &compressed_values,
-//     allocator,
-//     error_bound,
-// );
-// }
+    try compressSimPiece(
+        uncompressed_values[0..],
+        &compressed_values,
+        allocator,
+        error_bound,
+    );
+}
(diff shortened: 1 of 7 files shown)


src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        std.hash_map.default_max_load_percentage,
+    ), HashF64Context, std.hash_map.default_max_load_percentage),
+    allocator: mem.Allocator,
+) !void {
+    for (merged_segments_metadata.items) |segment_metadata| {
+        const interception: f64 = segment_metadata.interception;
+        const slope: f64 = (segment_metadata.lower_bound_slope +
+            segment_metadata.upper_bound_slope) / 2;
+
+        // Get or put the outer HashMap entry for the given interception
```
skejserjensen, line 409: Missing . at the end. Although maybe either delete the comment or extend it to state why it is done as it is clear a get put method is executed from the code.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        const hash_to_hash_result = try merged_segments_metadata_map.getOrPut(interception);
+        if (!hash_to_hash_result.found_existing) {
+            hash_to_hash_result.value_ptr.* = HashMap(
+                f64,
+                ArrayList(usize),
+                HashF64Context,
+                std.hash_map.default_max_load_percentage,
+            ).init(allocator);
+        }
+        // Get or put the inner HashMap entry for the given slope
```
skejserjensen, line 419: Missing . at the end. Although maybe either delete the comment or extend it to state why it is done as it is clear a get put method is executed from the code.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        const hash_to_array_result = try hash_to_hash_result.value_ptr.*.getOrPut(slope);
+        if (!hash_to_array_result.found_existing) {
+            hash_to_array_result.value_ptr.* = ArrayList(usize).init(allocator);
+        }
+        try hash_to_array_result.value_ptr.*.append(segment_metadata.start_time);
+    }
+}
+
+/// Sim-Piece Phase 4. Create from the `merged_segments_metadata_map` and byte array that can be
```
skejserjensen, line 428: Should and be an?

skejserjensen, line 428: Maybe add another sentence or two stating what the compressed representation is?

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+fn createCompressedRepresentation(
+    merged_segments_metadata_map: HashMap(f64, HashMap(
+        f64,
+        ArrayList(usize),
+        HashF64Context,
+        std.hash_map.default_max_load_percentage,
+    ), HashF64Context, std.hash_map.default_max_load_percentage),
+    compressed_values: *ArrayList(u8),
+) !void {
+    // Iterate over the hash to hash map.
```
skejserjensen, line 439: Maybe replace comment as it is clear that iteration is performed but it is not clear what the hash to hash map is.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    var hash_to_hash_iterator = merged_segments_metadata_map.iterator();
+    while (hash_to_hash_iterator.next()) |hash_to_hash_entry| {
+        const current_interception: f64 = hash_to_hash_entry.key_ptr.*;
+        // Append the `current_interception`.
```
skejserjensen, line 443: Maybe delete as it is clear from the code.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        try appendValue(f64, current_interception, compressed_values);
+
+        // Append the number of slopes hashed by the `current_interception`.
```
skejserjensen, line 446: Maybe update to also state what it means for slopes to hash to the same interception as it is not clear.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        try appendValue(usize, hash_to_hash_entry.value_ptr.*.count(), compressed_values);
+        var hash_to_array_iterator = hash_to_hash_entry.value_ptr.*.iterator();
+
+        while (hash_to_array_iterator.next()) |hash_to_array_entry| {
+            const current_slope = hash_to_array_entry.key_ptr.*;
+            // Append the `current_slope` that approximates the segment.
```
skejserjensen, line 452: Maybe delete as it is clear from the code.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+            try appendValue(f64, current_slope, compressed_values);
+
+            // Append the number of timestamps that the `current_slope` is hashing.
```
skejserjensen, line 455: Unclear what it means for a slope to be hashing, also maybe extend the comment to also state why this is done.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+            try appendValue(usize, hash_to_array_entry.value_ptr.*.items.len, compressed_values);
+            var previous_timestamp: usize = 0;
+            // Iterate over the timestamps and append them.
```
skejserjensen, line 458: Maybe delete this comment as it is clear from the code.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+            }
+        }
+    }
+}
+
+/// Quantizes the given `value` by the specified `error_bound`. This process ensures that
+/// the quantized value remains within the error bound of the original value. If the
+/// `error_bound` is equal to zero, the value is directly returned.
+fn quantize(value: f64, error_bound: f32) f64 {
+    if (error_bound != 0)
```
skejserjensen, line 471: Maybe add curly braces around the true block of the if expression for consistency and to minimize the chance of introducing bugs if the code is changed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        return @floor(value / error_bound) * error_bound;
+    return value;
+}
+
+/// Appends the `metadata` to the hash map `metadata_map`. The `allocator` to use for allocating
```
skejserjensen, line 476: Maybe replace "to use" with "is used" to make the sentence simpler to read.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    if (!get_result.found_existing) {
+        get_result.value_ptr.* = ArrayList(SegmentMetadata).init(allocator);
+    }
+    try get_result.value_ptr.*.append(metadata);
+}
+
+/// Returns a comparator function that compares SegmentMetadata by `lower_bound_slope`.
+fn compareMetadataBySlope() fn (void, SegmentMetadata, SegmentMetadata) bool {
+    return struct {
+        pub fn inner(_: void, metadata_one: SegmentMetadata, metadata_two: SegmentMetadata) bool {
```
skejserjensen, line 498: Unclear why this is not just a private function in the module scope instead of being part of a struct and then being returned as a function without the struct.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+            return metadata_one.lower_bound_slope < metadata_two.lower_bound_slope;
+        }
+    }.inner;
+}
+
+/// Returns a comparator function that compares SegmentMetadata by `start_time`.
+fn compareMetadataByStartTime() fn (void, SegmentMetadata, SegmentMetadata) bool {
+    return struct {
+        pub fn inner(_: void, metadata_one: SegmentMetadata, metadata_two: SegmentMetadata) bool {
```
skejserjensen, line 507: Unclear why this is not just a private function in the module scope instead of being part of a struct and then being returned as a function without the struct.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+            return metadata_one.start_time < metadata_two.start_time;
+        }
+    }.inner;
+}
+
+/// Append `value` of `type` determined at compile time to `compressed_values`.
+fn appendValue(comptime T: type, value: T, compressed_values: *std.ArrayList(u8)) !void {
+    // Compile-time type check
```
skejserjensen, line 515: Maybe delete this comment as it is clear from the code.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        f64,
+        f64,
+        HashF64Context,
+        std.hash_map.default_max_load_percentage,
+    ).init(allocator);
+    defer f64_hash_map.deinit();
+    var rnd = std.rand.DefaultPrng.init(@as(u64, @bitCast(std.time.milliTimestamp())));
+
+    // Even a small deviation like this should not impact the ability to hash the numbers.
+    const deviation_margin = 1e-16;
```
skejserjensen, line 541: He purposely of the margin is unclear, maybe extend the comment.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    for (0..100) |_| {
+        const rand_number = rnd.random().float(f64) - 0.5;
+        try f64_hash_map.put(rand_number, rand_number);
+        try f64_hash_map.put(rand_number - deviation_margin, rand_number - deviation_margin);
+        try f64_hash_map.put(rand_number + deviation_margin, rand_number + deviation_margin);
+    }
+
+    // All elements are expected to be hashed independently, without collisions.
+    try testing.expectEqual(300, f64_hash_map.count());
```
skejserjensen, line 550: How does `count()` if the values does not collide? If this is done maybe extend the comment so it explains how, if it is not done maybe delete the comment or revise it.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+            try testing.expectEqual(i, item.start_time);
+            try testing.expectEqual(entry.key_ptr.*, item.lower_bound_slope);
+        }
+    }
+}
+
+test "sim-piece can compress, decompress and merge many segments with non-zero error bound" {
+    const allocator = testing.allocator;
+
+    var list_values = ArrayList(f64).init(allocator);
```
skejserjensen, line 622: The purpose of `list_values` is unclear for most of the test. Maybe rename to `uncompressed_values` and convert to slice when calling `compress()` so `list_values` can be removed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    defer list_values.deinit();
+    var compressed_values = ArrayList(u8).init(allocator);
+    defer compressed_values.deinit();
+    var decompressed_values = ArrayList(f64).init(allocator);
+    defer decompressed_values.deinit();
+    var rnd = std.rand.DefaultPrng.init(@as(u64, @bitCast(std.time.milliTimestamp())));
+
+    const error_bound: f32 = @floor(rnd.random().float(f32) * 100) / 10 + 0.1;
+
+    for (0..200) |_| {
```
skejserjensen, line 632: Unclear what render many merged segments mean.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        // Generate floating points between numbers between 0 and 10. This will render many
```
skejserjensen, line 633: Unclear what render many merged segments mean.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        },
+        f32 => {
+            const value_as_bytes: [4]u8 = @bitCast(value);
+            try compressed_values.appendSlice(value_as_bytes[0..]);
+        },
+        else => @compileError("Unsupported type for append value function"),
+    }
+}
+
+test "f64 context can hash" {
```
skejserjensen, line 529: Since the tests in this file use random data, maybe implement them using `tester.zig` and extend it with functions for generating data with the required structure if such functions does not already exist. Since most of our tests for TerseTS consist of generating data, compress, decompress, and check, implementing functionality for this in a single place instead of in each test seems like a good way to significantly reduce code duplication and the amount of effort required to write tests.

skejserjensen (changes requested):
As a general comment, maybe add, remove, and update the comments and doc comments so they explain anything that it is important that the reader understands about the code, but which cannot be understood or is hard to understand from reading the code.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Implementation of the Sim-Piece algorithm from the paper
+//! "Xenophon Kitsios, Panagiotis Liakos, Katia Papakonstantinopoulou, and Yannis Kotidis.
+//! Sim-Piece: Highly Accurate Piecewise Linear Approximation through Similar Segment Merging.
+//! Proc. VLDB Endow. 16, 8 2023.
+//! https://doi.org/10.14778/3594512.3594521".
+//! The implementation is partially based on the authors implementation at
+//! https://github.com/xkitsios/Sim-Piece (accessed on 20-06-24). Few changes were made to support
```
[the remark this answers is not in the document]
cmcuza: The original implementation does not take care of it and it will even return a division by zero error. The authors assumed zero is not given.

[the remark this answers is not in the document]
skejserjensen: Would an implementation that returns an `UnsupportedErrorBound` error if the error bound is not a positive value and otherwise match the original implementation not be a safe implementation of the compression method in the paper? Maybe we should rename `NegativeErrorBound` to `UnsupportedErrorBound` to not create two different errors for an incorrect error bound.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+const testing = std.testing;
+const ArrayList = std.ArrayList;
+const HashMap = std.HashMap;
+
+const tersets = @import("../tersets.zig");
+const Error = tersets.Error;
+const DiscretePoint = tersets.DiscretePoint;
+
+/// `SegmentMetadata` stores the information about an approximated segment during the execution
+/// of Sim-Piece. It stores the starting time of the segment in `start_time`, the slopes of
```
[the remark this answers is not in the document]
cmcuza: Fixed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+/// the upper and lower bounds that constraint the linear approximation in that segment and the
+/// `interception` point used to create the linear function approximation.
+const SegmentMetadata = struct {
+    start_time: usize,
+    interception: f64,
+    upper_bound_slope: f64,
+    lower_bound_slope: f64,
+};
+
+/// `HashF64Context` is a struct providing context for hashing and comparing `f64` values.
```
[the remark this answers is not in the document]
cmcuza: I don't know what `Context` means either. However, in https://ziglang.org/documentation/master/std/#std.hash_map.HashMap they called context. I added more information about why is needed for HashMap.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+/// the result to `compressed_values`. The `allocator` is used for memory allocation of intermediate
+/// data structures. If an error occurs, it is returned.
+pub fn compressSimPiece(
+    uncompressed_values: []const f64,
+    compressed_values: *ArrayList(u8),
+    allocator: mem.Allocator,
+    error_bound: f32,
+) Error!void {
+    if (error_bound == 0.0) {
+        std.debug.print("\nWarning: Sim-Piece does not support lossless compression. ", .{});
```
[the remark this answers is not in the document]
cmcuza: Ok, I found this solution to allow a zero error bound. But, I now is remove together with the extra code.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        std.debug.print("Expect deviations from the original data of at most 1e-6.\n", .{});
+    }
+    // Sim-Piece Phase 1: Compute the segments metadata.
```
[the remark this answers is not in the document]
cmcuza: Well, each function already has a description of what is doing. I can repeat information but it seems unnecessary. I tried to increase the comment without repeat information.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        error_bound,
+    );
+
+    // Sim-Piece Phase 2: Merge the segments metadata.
+    var merged_segments_metadata = ArrayList(SegmentMetadata).init(allocator);
+    defer merged_segments_metadata.deinit();
+    try mergeSegmentsMetadata(segments_metadata, &merged_segments_metadata, allocator);
+
+    // Sim-Piece Phase 3: Compute segment metadata hash map.
+    var merged_segments_metadata_map = HashMap(
```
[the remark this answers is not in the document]
cmcuza: Nice, the code looks much simpler now.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        HashMap(
+            f64,
+            ArrayList(usize),
+            HashF64Context,
+            std.hash_map.default_max_load_percentage,
+        ),
+        HashF64Context,
+        std.hash_map.default_max_load_percentage,
+    ).init(allocator);
+    defer {
```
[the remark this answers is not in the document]
cmcuza: Fixed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    }
+    try computeSegmentsMetadataHashMap(
+        merged_segments_metadata,
+        &merged_segments_metadata_map,
+        allocator,
+    );
+
+    // Sim-Piece Phase 4: Create the final compressed representation and store in compressed values.
+    try createCompressedRepresentation(merged_segments_metadata_map, compressed_values);
+    // The last timestamp needs to be stored in the compressed representaiton as well.
```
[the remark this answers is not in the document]
cmcuza: Fixed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+}
+
+/// Decompress `compressed_values` produced by "Sim-Piece" and write the result to
+/// `decompressed_values`. If an error occurs it is returned.
+pub fn decompress(
+    compressed_values: []const u8,
+    decompressed_values: *ArrayList(f64),
+    allocator: mem.Allocator,
+) Error!void {
+    // The compressed representation of Sim-Piece is of variable length. We cannot assert the len
```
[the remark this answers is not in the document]
cmcuza: Yes. In the previous decompression method, we added an `assert` on the length of the compressed representation. However, now is of variable length.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    );
+
+    // Sim-Piece Phase 4: Create the final compressed representation and store in compressed values.
+    try createCompressedRepresentation(merged_segments_metadata_map, compressed_values);
+    // The last timestamp needs to be stored in the compressed representaiton as well.
+    try appendValue(usize, uncompressed_values.len, compressed_values);
+}
+
+/// Decompress `compressed_values` produced by "Sim-Piece" and write the result to
+/// `decompressed_values`. If an error occurs it is returned.
```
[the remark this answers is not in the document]
cmcuza: Fixed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    decompressed_values: *ArrayList(f64),
+    allocator: mem.Allocator,
+) Error!void {
+    // The compressed representation of Sim-Piece is of variable length. We cannot assert the len
+    // of the compressed representation to be equal to any specific number.
+    var segments_metadata = ArrayList(SegmentMetadata).init(allocator);
+    defer segments_metadata.deinit();
+    const compressed_lines_and_index = mem.bytesAsSlice(f64, compressed_values);
+
+    var general_index: usize = 0;
```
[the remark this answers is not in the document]
cmcuza: A remnant of a previous version with various indices. Anyways, it is now changed to `compressed_index`.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        while (current_timestamp < next_metadata_start_time) : (current_timestamp += 1) {
+            const approximated_value = current_metadata.upper_bound_slope * @as(f64, @floatFromInt(
+                current_timestamp - current_metadata.start_time,
+            )) + current_metadata.interception;
+            try decompressed_values.append(approximated_value);
+        }
+    }
+
+    const current_metadata = segments_metadata.getLast();
+    while (current_timestamp < last_timestamp) : (current_timestamp += 1) {
```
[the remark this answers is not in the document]
cmcuza: Fixed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        const approximated_value = current_metadata.upper_bound_slope * @as(f64, @floatFromInt(
+            current_timestamp - current_metadata.start_time,
+        )) + current_metadata.interception;
+        try decompressed_values.append(approximated_value);
+    }
+}
+
+/// Sim-Piece Phase 1: Compute the `segments_metadata` for each segment that can be approximated
```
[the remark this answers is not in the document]
cmcuza: The first one, fixed.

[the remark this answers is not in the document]
cmcuza: Fixed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+/// by a linear function within the `error_bound` from the `uncompressed_values`.
+fn computeSegmentsMetadata(
+    uncompressed_values: []const f64,
+    segments_metadata: *ArrayList(SegmentMetadata),
+    error_bound: f32,
+) !void {
+    // Adjust the error bound to avoid exceeding it during decompression.
+    const adjusted_error_bound = if (error_bound > 0)
```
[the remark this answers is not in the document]
cmcuza: I could do that, but since this function needs both I will have to pass them as parameters. It think it is better to leave like this since nowhere else the adjusted error bound is computed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    else
+        error_bound;
+
+    var upper_bound_slope: f64 = math.floatMax(f64);
+    var lower_bound_slope: f64 = -math.floatMax(f64);
+
+    // Initialize the `start_point` with the first uncompressed value.
+    var start_point: DiscretePoint = .{ .time = 0, .value = uncompressed_values[0] };
+
+    var quantized_interception = quantize(uncompressed_values[0], error_bound) +
```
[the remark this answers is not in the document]
cmcuza: Fixed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        tersets.ErrorBoundMargin;
+
+    // First point already part of `current_segment`, next point is at index one.
```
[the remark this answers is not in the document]
cmcuza: Fixed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    allocator: mem.Allocator,
+) Error!void {
+    // The compressed representation of Sim-Piece is of variable length. We cannot assert the len
+    // of the compressed representation to be equal to any specific number.
+    var segments_metadata = ArrayList(SegmentMetadata).init(allocator);
+    defer segments_metadata.deinit();
+    const compressed_lines_and_index = mem.bytesAsSlice(f64, compressed_values);
+
+    var general_index: usize = 0;
+    while (general_index < compressed_lines_and_index.len - 1) {
```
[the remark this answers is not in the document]
cmcuza: You are right, in this case, the for loop can be used. I will check everywhere and fix it accordingly.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+
+            if (end_point.value + adjusted_error_bound < upper_limit)
+                upper_bound_slope = @max(new_upper_bound_slope, lower_bound_slope);
+            if (end_point.value - adjusted_error_bound > lower_limit)
+                lower_bound_slope = @min(new_lower_bound_slope, upper_bound_slope);
+        }
+    }
+
+    const segment_size = current_timestamp - start_point.time;
+    if (segment_size >= 1) {
```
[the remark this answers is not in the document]
cmcuza: Sure.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        try segments_metadata.append(.{
+            .start_time = start_point.time,
+            .interception = quantized_interception,
+            .upper_bound_slope = upper_bound_slope,
+            .lower_bound_slope = lower_bound_slope,
+        });
+    }
+}
+
+/// Sim-Piece Phase 2. Merge the elements in `segments_metadata` based on Alg. 2 and store the
```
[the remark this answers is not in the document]
cmcuza: Done. Also, added a small description as suggested before, so the user does not need to look into the paper.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    var timestamps_array = ArrayList(usize).init(allocator);
+    defer timestamps_array.deinit();
+
+    var segments_metadata_map = HashMap(
+        f64,
+        ArrayList(SegmentMetadata),
+        HashF64Context,
+        std.hash_map.default_max_load_percentage,
+    ).init(allocator);
+    defer {
```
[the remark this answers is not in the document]
cmcuza: Done/

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+        var iterator = segments_metadata_map.iterator();
+        while (iterator.next()) |entry| {
+            entry.value_ptr.*.deinit();
+        }
+        segments_metadata_map.deinit();
+    }
+
+    // Iteratively populate `segments_metadata_map` based on the quantized interception values to
+    // find the common interception between the segments. Segments falling in the same interception
```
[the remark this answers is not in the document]
cmcuza: Much better this way.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    // can potentially be merged.
```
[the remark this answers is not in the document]
cmcuza: Done.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+            segment_metadata,
+            allocator,
+        );
+    }
+
+    var iterator = segments_metadata_map.iterator();
+    while (iterator.next()) |entry| {
+        const metadata_array = entry.value_ptr.*;
+
+        // Sort in asc order based on the lower bound's slope. Alg 2. Line 5.
```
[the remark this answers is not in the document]
cmcuza: I am explanation can be more elaborate but I want to overdo it either. I will wait for your opinion with the new comment.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+                    .start_time = 0.0,
+                    .interception = current_metadata.interception,
+                    .lower_bound_slope = current_metadata.lower_bound_slope,
+                    .upper_bound_slope = current_metadata.upper_bound_slope,
+                };
+                try timestamps_array.append(current_metadata.start_time);
+            }
+        }
+
+        if (timestamps_array.items.len > 0) {
```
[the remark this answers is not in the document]
cmcuza: Good catch. Fixed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+                    .interception = merge_metadata.interception,
+                    .lower_bound_slope = merge_metadata.lower_bound_slope,
+                    .upper_bound_slope = merge_metadata.upper_bound_slope,
+                });
+            }
+            timestamps_array.clearRetainingCapacity();
+        }
+    }
+
+    // This step is needed since the timestamp order is lost due to the hashmap.
```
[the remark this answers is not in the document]
cmcuza: Fixed.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    mem.sort(
+        SegmentMetadata,
+        merged_segments_metadata.items,
+        {},
+        compareMetadataByStartTime(),
+    );
+}
+
+/// Sim-Piece Phase 3. Compute a hash map from interception points in `merged_segments_metadata`
```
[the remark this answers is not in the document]
cmcuza: I agree, the word "compute" can be unclear here. Another option could be "populate", but I am not sure, and it is also unclear. Seems like any possible option can receive the same comment. Thus, I leave it as it is, but improved the comment as suggested.

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+/// to a hash map from the approximation slope to an array list of timestamps and store in
```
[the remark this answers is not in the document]
cmcuza: Changed to "return it in". I think is clearer.

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
cmcuza: I extended a bit to show the logic of the HashMaps.

src/functional/sim_piece.zig
cmcuza: My doc was wrong. I fixed it.

src/functional/sim_piece.zig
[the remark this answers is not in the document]
cmcuza: I changed it back again to "store it in" for consistency with other docs.

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
cmcuza: Changed to // Iterate over the outer HashMap.

src/functional/sim_piece.zig
cmcuza: Done.

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
cmcuza: Done.

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
cmcuza: I think for consistency I will leave it. Now is improved.

src/functional/sim_piece.zig
cmcuza: Done.

src/functional/sim_piece.zig
cmcuza: I did this example based on:
https://zig.guide/standard-library/sorting/
https://ziglang.org/documentation/master/std/#std.sort.asc
I adapted it to our use case. However, I am not even sure how it works. I will try a few things, if I cannot get it to work, I will leave it as it is.

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
cmcuza: I think the comment now explains the logic.

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
cmcuza: This was a good idea. A lot of duplicated code is gone.

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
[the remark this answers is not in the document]
cmcuza: Fixed.

src/functional/sim_piece.zig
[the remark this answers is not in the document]
skejserjensen: Could [this](https://github.com/cmcuza/TerseTS/blob/dev/add-sim-piece/src/functional/sim_piece.zig#L133) `while` loop not also be simplified to a `for` loop?

src/functional/sim_piece.zig
```diff
@@ -0,0 +1,765 @@
...
+    for (segments_metadata.items) |segment_metadata| {
+        try appendSegmentMetadata(
+            &segments_metadata_map,
+            segment_metadata,
+            allocator,
+        );
+    }
+
+    var iterator = segments_metadata_map.iterator();
+    while (iterator.next()) |entry| {
```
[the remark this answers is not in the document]
skejserjensen: Ok, if `while` loops are generally used for iterators in Zig we should probably stick to that. Good catch.

src/functional/sim_piece.zig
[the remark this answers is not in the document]
skejserjensen: "This allows to find..." makes it sound like it could not be done without the sorting, is this correct or is the process just more efficient if they are sorted? If it is just more efficient maybe replace it with "The makes it more efficient to find ..." or if it is required maybe replace it with "This makes it possible to find ..." to make it very clear that this is a requirement of the algorithm.

src/functional/sim_piece.zig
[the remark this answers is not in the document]
skejserjensen: In this case I would use "populate" in both the comment and in the function name as the function seems to populate `merged_segments_metadata_map` based on the elements in `merged_segments_metadata` and thus I think populate better describes what the function does than compute. However, if the function was calculating a single result from `merged_segments_metadata` I would probably use compute instead of populate.

src/functional/sim_piece.zig
skejserjensen: Of course it depends on how annoyingly pedantic the reviewer is :grin:

In my opinion there is very little benefit to not write the curly braces and if they are always included the code becomes more consistent and it becomes more clear what lines of code are included as part of the branch. Others seems to have come to the same conclusion as the C++ style guides I am familiar with also requires that curly braces are always included except for very specific situations, e.g., see [Google's C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Formatting_Looping_Branching).

src/functional/sim_piece.zig
skejserjensen: See this [comment](https://github.com/cmcuza/TerseTS/pull/43#discussion_r1667688424).

src/functional/sim_piece.zig
skejserjensen: It may just be me that is missing something, but it is still not clear to me how the use of `count()` can be used to check for [hash collisions](https://en.wikipedia.org/wiki/Hash_table#Collision_resolution). Do you mean that the values should all be stored as separate key-value pairs in the hash map and not that hash collisions have occurred?

src/functional/sim_piece.zig
skejserjensen: Is `[0..]` not only needed if a sub-slice is needed? Does the simpler `uncompressed_values.items` or `&uncompressed_values.items` not work?

src/functional/sim_piece.zig
skejserjensen: Are the values guaranteed to create many merged segments? If so maybe replace "must" with "will".

skejserjensen (changes requested):
Look good, there are only a few minor comments left, mainly just things I do not understand.

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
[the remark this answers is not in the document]
cmcuza: No, I need to increase the index at different rates depending what I find inside the loop.

src/functional/sim_piece.zig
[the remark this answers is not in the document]
cmcuza: The truth is that finding the merging segments without sorting is quite complicated, so when I say allows, I mean to enable the merging process.

src/functional/sim_piece.zig
cmcuza: Yes, that is exactly what I mean. Changed the comment to make it even clearer.

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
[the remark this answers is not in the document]
cmcuza: Fixed. Changed to populate.

src/functional/sim_piece.zig
cmcuza: Great. Fixed!

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
skejserjensen: Very minor, but maybe remove "Returns a" from the comment as the function now is the comparator and returns a `bool`.

src/functional/sim_piece.zig
skejserjensen: Very minor, but maybe remove "Returns a" from the comment as the function now is the comparator and returns a `bool`.

skejserjensen (changes requested):
Very minor, but seems the doc comment for the comparator functions were not updated

src/functional/sim_piece.zig
cmcuza: Fixed.

src/functional/sim_piece.zig
cmcuza: Fixed.

skejserjensen (approved):

cmcuza:
The PR is still blocked.

skejserjensen (approved):

skejserjensen (approved):

skejserjensen:
> The PR is still blocked.

Seems

> The PR is still blocked.

It seems to be a GitHub issue as not commits have been made since I approved the PR and I have tried to approve it again and request a review from myself and approve it again without any effect. Maybe disable the requirement that approval must be done on the latest commit in branch predication, merge this PR, and then enable it again.

cmcuza:
> > The PR is still blocked.
> 
> It seems to be a GitHub issue as not commits have been made since I approved the PR and I have tried to approve it again and request a review from myself and approve it again without any effect. Maybe disable the requirement that approval must be done on the latest commit in branch predication, merge this PR, and then enable it again.

The requirement was disabled to continue with the PR and enabled back again.

Issue #45, opened 2024-06-21
Implement Maximum Error Histogram Algorithm


cmcuza:
This issue is for implementing the methods described in the paper: Space Efficient Streaming Algorithms for the Maximum Error Histogram. There, two different algorithms are proposed: piecewise constant histogram (PCH) and piecewise linear histogram (PLH). Both can be considered lossy time series compression techniques and will be valuable assets to TerseTS.

cmcuza:
PWLH is still missing. This I reopen this issue.

PR #51, opened 2024-10-29
Add Minimum-Area Bounding Rectangle

cmcuza:
This PR updates the convex hull data structure to compute the Minimum-Area Bounding Rectangle (MABR) based on the Rotating Calipers algorithm [1]. With this algorithm, we can then compute the linear function that bisects the MABR following [2], which is related to #45. Besides the new function, we reformed a `capi.zig` and `tersets.zig` to have access to the LinearFunction structure and the `tersets.Error`. I have added a few three tests, but is quite hard to generate a bigger test with random numbers since the expected output is hard to compute manually. The two hardcoded tests are working well.

[1] Shamos, Michael (1978). "Computational Geometry" (PDF). Yale University. pp. 76–81.
[2] https://doi.org/10.1109/ICDE.2007.368961.


src/functional/geometry/convex_hull.zig
```diff
@@ -66,22 +72,127 @@ pub const ConvexHull = struct {
...
         }
-        return self.lower_hull.items[0 .. len - 1];
+        return self.lower_hull.items[0 .. lower_hull_len - 1];
+    }
+
+    /// Computes the `LinearFunction` with minimum square error based on the Minimum-Area Bounding
+    /// Rectangle (MABR) of a Convex Hull. The MABR is computed using the Rotating Calipers
+    /// algorithm described in the paper:
+    /// Shamos, Michael (1978). "Computational Geometry" (PDF). Yale University. pp. 76–81.
+    /// Our implementation is based mostly on the description found in:
```
skejserjensen, line 97: Unclear who Our is in this context, maybe replace with This.

src/functional/geometry/convex_hull.zig
```diff
@@ -66,22 +72,127 @@ pub const ConvexHull = struct {
...
+    /// https://en.wikipedia.org/wiki/Rotating_calipers.
+    pub fn computeMABRLinearFunction(self: *ConvexHull) Error!LinearFunction {
+        const hullPoints: ArrayList(DiscretePoint) = try self.getAllPoints();
+        defer hullPoints.deinit();
+        const convex_hull_len: usize = hullPoints.items.len;
+
+        // One point in the ConvexHull, the line has `slope=0`.
```
skejserjensen, line 104: Unclear what One point and Two Point means in this context, maybe expand their explanation or replace them with something like the "first" and "last" point of the linear function.

skejserjensen, line 104: Unclear what slope equals zero means in this context, is it simply because there is only one point for the function? Maybe expand their explanation.

src/functional/geometry/convex_hull.zig
```diff
@@ -66,22 +72,127 @@ pub const ConvexHull = struct {
...
+            const second_point = hullPoints.items[1];
+            const delta_time: f80 = @floatFromInt(second_point.time - first_point.time);
+            const slope = (second_point.value - first_point.value) / delta_time;
+            const intercept_value: f80 = first_point.value - slope * @as(f80, @floatFromInt(first_point.time));
+            return LinearFunction{
+                .slope = slope,
+                .intercept = intercept_value,
+            };
+        }
+        // Initialize variables to track the minimum area and corresponding parameters
```
skejserjensen, line 121: From here some comments are missing punctuation at the end.

src/functional/geometry/convex_hull.zig
```diff
@@ -37,12 +41,14 @@ const Turn = enum(i8) { right, left, collinear };
 pub const ConvexHull = struct {
```
skejserjensen, line 41: The instance variables are missing doc comments.

src/functional/geometry/convex_hull.zig
```diff
@@ -66,22 +72,127 @@ pub const ConvexHull = struct {
...
+        // Iterate over each edge of the convex hull
+        for (0..convex_hull_len) |i| {
+            // Get the current edge (segment) of the convex hull
+            const first_point: DiscretePoint = hullPoints.items[i];
+            const second_point: DiscretePoint = hullPoints.items[@mod(i + 1, convex_hull_len)];
+            const segment = Segment{ .start_point = first_point, .end_point = second_point };
+
+            // Compute the angle between the current edge and the x-axis
+            const angle = angleToXAxis(segment);
+            // Initialize min/max values for rotated points
```
skejserjensen, line 135: Maybe add new line before the comment as this seems to be generally done.

src/functional/geometry/convex_hull.zig
```diff
@@ -66,22 +72,127 @@ pub const ConvexHull = struct {
...
+            }
+        }
+
+        return LinearFunction{ .slope = min_slope, .intercept = min_intercept };
+    }
+
+    /// Helper function to get all points as an `ArrayList(DiscretePoint)` from the ConvexHull.
+    /// If the `ConvexHull` is empty, an empty ArrayList is returned.
+    fn getAllPoints(self: *ConvexHull) !ArrayList(DiscretePoint) {
+        var hull: ArrayList(DiscretePoint) = ArrayList(DiscretePoint).init(self.allocator);
```
skejserjensen, line 187: Seems costly to allocate memory for all of the points in the hull and then copy them to the `ArrayList`. Maybe create some sort of iterator that can be used to iterate through the hull without copying it under the assumption that it is not changed during iteration? If this is not possible maybe allocate enough memory for the hull instead of allocating memory multiple times by growing the `ArrayList` as the hulls size is known.

src/functional/geometry/convex_hull.zig
```diff
@@ -103,6 +214,15 @@ pub const ConvexHull = struct {
             try hull.append(point);
         }
     }
+
+    /// Returns the length of the `ConvexHull`. Since the first and final element are repeated,
```
skejserjensen, line 218: Maybe make the part of the doc comment about subtracting two a comment for the last line of code as it seems to be an implementation detail and thus not information the user needs to consider.

src/functional/geometry/convex_hull.zig
```diff
@@ -204,3 +345,77 @@ test "incremental convex hull random elements" {
         try testing.expectEqual(turn, Turn.left);
     }
 }
+
+test "Compute MABR LinearFunction for known ConvexHull one" {
```
skejserjensen, line 349: Maybe add a space before Hull for consistency.

src/functional/geometry/convex_hull.zig
```diff
@@ -204,3 +345,77 @@ test "incremental convex hull random elements" {
...
+
+    // Initialize the convex hull and add points to it.
+    var convex_hull = try ConvexHull.init(allocator);
+    defer convex_hull.deinit();
+
+    for (points) |point| {
+        try convex_hull.addPoint(point);
+    }
+
+    // Calculate the minimum-area bounding rectangle (MABR).
```
skejserjensen, line 369: Seems only the function is returned and not the MABR, maybe update the comment to reflect this.

src/functional/geometry/convex_hull.zig
```diff
@@ -204,3 +345,77 @@ test "incremental convex hull random elements" {
...
+    const mabr_linear_function = try convex_hull.computeMABRLinearFunction();
+
+    // Known rectangle area and known LinearFunction y=x+1
```
skejserjensen, line 372: Unclear what known rectangle area means in this context, maybe remove it.

skejserjensen, line 372: Unclear what y=x+1 means in this context, maybe expand it.

src/functional/geometry/convex_hull.zig
```diff
@@ -204,3 +345,77 @@ test "incremental convex hull random elements" {
...
+    var rnd = std.rand.DefaultPrng.init(0);
+
+    var convex_hull = try ConvexHull.init(allocator);
+    defer convex_hull.deinit();
+
+    for (0..num_points) |i| {
+        try convex_hull.addPoint(.{ .time = i, .value = rnd.random().float(f64) });
+    }
+
+    // Calculate the minimum-area bounding rectangle (MABR). If dont have way to check if the
```
skejserjensen, line 416: Unclear what if don't mean, maybe update the comment.

src/functional/geometry/convex_hull.zig
```diff
@@ -204,3 +345,77 @@ test "incremental convex hull random elements" {
...
+    defer convex_hull.deinit();
+
+    for (points) |point| {
+        try convex_hull.addPoint(point);
+    }
+
+    // Calculate the minimum-area bounding rectangle (MABR).
+    const mabr_linear_function = try convex_hull.computeMABRLinearFunction();
+
+    // Known rectangle area and known LinearFunction y=x+1
```
skejserjensen, line 399: Seems only the function is returned and not the MABR, maybe update the comment to reflect this.

skejserjensen, line 399: Unclear what y=x+1 means in his context, maybe update the comment.

src/functional/geometry/convex_hull.zig
```diff
@@ -204,3 +345,77 @@ test "incremental convex hull random elements" {
...
+
+    for (0..num_points) |i| {
+        try convex_hull.addPoint(.{ .time = i, .value = rnd.random().float(f64) });
+    }
+
+    // Calculate the minimum-area bounding rectangle (MABR). If dont have way to check if the
+    // result is correct but we can check that no error has happened.
+    const mabr_linear_function = try convex_hull.computeMABRLinearFunction();
+
+    try testing.expect(mabr_linear_function.intercept < 1.0);
```
skejserjensen, line 420: Unclear how it is known that the intercept is below zero and why it is important to check.

src/tersets.zig
```diff
@@ -53,6 +54,12 @@ pub const Segment = struct {
     end_point: DiscretePoint,
 };
 
+/// Linear function of the form y = slope*x+intercept. It uses f80 for numerical stability.
```
skejserjensen, line 57: As mentioned in #50, maybe create a shared folder containing all helper types, functions, and methods. See [this comment](https://github.com/cmcuza/TerseTS/pull/50#discussion_r1827024730) for more details.

skejserjensen (changes requested):

src/functional/geometry/convex_hull.zig
cmcuza: fixed.

src/functional/geometry/convex_hull.zig
cmcuza: fixed.

cmcuza: fixed.

src/functional/geometry/convex_hull.zig
cmcuza: fixed.

src/functional/geometry/convex_hull.zig
cmcuza: fixed.

src/functional/geometry/convex_hull.zig
cmcuza: I tried this before but did not manage due to repeated first and last element in the lower and upper hulls.

src/functional/geometry/convex_hull.zig
cmcuza: fixed

src/functional/geometry/convex_hull.zig
cmcuza: fixed

src/functional/geometry/convex_hull.zig
cmcuza: An old comment. Now it is fixed.

cmcuza: fixed.

src/functional/geometry/convex_hull.zig
cmcuza: I thought to use this test to check that no errors occur with a random set of numbers. Nevertheless, since all points are between 0-1 then the intercept must be less than 1. 
I fixed the comment to make it less intricate.

src/functional/geometry/convex_hull.zig
cmcuza: fixed.

cmcuza: fixed.

src/functional/geometry/convex_hull.zig
cmcuza: I thought to use this test to check that no errors occur with a random set of numbers. Nevertheless, since all points are between 0-1 then the intercept must be less than 1. This is because the linear function is the height bisector and goes through the midpoint of the rectangle side which has as lower and upper bounds 0 and 1.

src/tersets.zig
cmcuza: That makes a lot of sense. Then, I will keep this PR open, even if all the issues are solved. Once I merge the other PR, I will ask for another review for this one.

cmcuza: Or if you want to accept this PR first and later I move the Convex Hull to another folder, we can do that as well.

cmcuza: In the other PR, I address this problem, let me know if you accept that idea and if I should reproduce in this one.

src/functional/geometry/convex_hull.zig
skejserjensen: It is still unclear why a [`initCapacity()`](https://ziglang.org/documentation/master/std/#std.array_list.ArrayListAligned.initCapacity) is not used to allocate the required amount of memory once, instead of multiple times using `init()`, as it seems the number of points can be efficiently computed from the length of lower and upper pats of `ConvexHull`.

src/tersets.zig
skejserjensen: I suggest you create an issue to make this change after the two PRs have been merged. Since GitHub does not differentiate between code that has been moved and code that has been changed, I find it very hard to review a PR that does both.

skejserjensen (changes requested):
See the new comments to the conversations.

src/tersets.zig
cmcuza: Ummm I see, ok I will do that.

src/functional/geometry/convex_hull.zig
cmcuza: Umm, I did not know that function. Fixed.

src/tersets.zig
skejserjensen: Ok, I have approved the PR as  this is the last unresolved conversation and I have no more comments. Maybe post a link to the issue in this conversation when it is created and link to this conversation in the issue so the connection is clear.

skejserjensen (approved):

src/tersets.zig
cmcuza: Done!

cmcuza: I will resolve the conversation :)

PR #61, opened 2025-01-05
Python Package

### LINKED ISSUE #52: Create PyPI package for Python
To simplify installation and use of TerseTS in Python, a PyPI package should be created. As Zig is available as a [PyPI package](https://github.com/ziglang/zig-pypi) it should be possible to create a package that will download the correct Python `wheel` if one exists, and if not install Zig and build the package from source. Based on a quick search, it seems at least a few example packages written in Zig are available such as [zaml](https://github.com/adamserafini/zaml/tree/main) and [python_zig](https://github.com/Lucifer-02/python_zig). But it would be preferable if mature PyPI packages written in Zig can be found so we can learn from them.

skejserjensen:
This fixes #52 by making the Python bindings a [Python package](https://packaging.python.org/en/latest/) that can be uploaded to PyPi. The package uses [`setuptools`](https://setuptools.pypa.io/en/latest/) as the build backend and extends the `sdist` and `build_ext` commands to compile and bundle the Zig code. The extensions are implemented in a `setup.py` file and is needed as `setuptools` assume extension are written in [C or C++](https://setuptools.pypa.io/en/latest/userguide/ext_modules.html). The package can be constructed with `python -m build` and installed with `python -m pip install` as [recommended](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/). The package has been tested on Arch Linux, x86_64 macOS, arm64 macOS, and Microsoft Windows 11. In addition to reviewing the code, it should also be verified that the package can be created on the most common platforms, that it can be installed on them, that `tersets` can be imported on them with the package installed, and that the bindings can load library compiled in the repository on them to keep development simple.


bindings/python/pyproject.toml
```diff
@@ -15,7 +15,8 @@
 
 [project]
 name = "TerseTS"
-description = "Python interface for the TerseTS time series compression library"
+description = "TerseTS Python bindings"
+readme = "README.md"
 version = "0.0.1"
 license = { text="Apache-2.0" }
 authors = [
```
cmcuza, line 22: Can we update my email as part of this PR? If so, please update with muniz.cuza@tu-berlin.de

bindings/python/setup.py
```diff
@@ -0,0 +1,73 @@
...
+        # CPython and it simplifies the code for loading the library in Python.
+        filename = super().get_ext_filename(ext_name)
+        start = filename.find(".")
+        end = filename.rfind(".")
+        return filename[:start] + filename[end:]
+
+
+setup(
+    packages=find_packages(),
+    ext_modules=[Extension("tersets", sources=["src/capi.zig"])],
```
cmcuza, line 71: When trying to install `tersets` using `python -m pip install .` inside the `binding/python/` directory, `src/capi.zig` is not found. However, if this is changed to `../../src/capi.zig`, then `python -m build` does not work. This was the problem you had before when commenting on the issue. That means that the correct way to install `tersets` from the repository is by first building and creating the wheel distribution, for example: `dist/TerseTS-0.0.1-cp311-cp311-win_amd64.whl`, and later using `pip install dist/TerseTS-....whl`. Should we document this in the `bindings/python/README.md`?

If we do so, should we also specify that they need to install the package `build` and, of course, install `zig` locally?

bindings/python/setup.py
```diff
@@ -0,0 +1,73 @@
...
+                ext.sources[0],
+                f"-femit-bin={self.get_ext_fullpath(ext.name)}",
+                "-dynamic",
+                "-O",
+                "ReleaseFast",
+             ],
+         )
+
+        # Zig generates files that are not needed and can be removed.
+        if sys.platform == "darwin" or sys.platform == "linux":
```
cmcuza, line 54: Consider changing the code to this:

```python
if sys.platform in ("darwin", "linux"):
            try:
                os.remove(self.get_ext_fullpath(ext.name) + ".o")
            except FileNotFoundError:
                pass
  ```

File: bindings/python/setup.py [hunk 4]
```diff
@@ -0,0 +1,73 @@
...
+            os.remove(self.get_ext_fullpath(ext.name) + ".o")
+        elif sys.platform == "win32":
+            for name in ["capi.lib", "tersets.pdb", "tersets.pyd.obj"]:
+                os.remove(self.build_lib + "\\" + name)
```
cmcuza, line 58: Consider changing the code to this:
```python
unwanted = os.path.join(self.build_lib, name)
if os.path.exists(unwanted):
    os.remove(unwanted)
    ```
I think using `os.path.join` is safer.

File: bindings/python/setup.py [hunk 5]
```diff
@@ -0,0 +1,73 @@
...
+import shutil
+
+from setuptools import setup, find_packages, Extension
+from setuptools.command.sdist import sdist
+from setuptools.command.build_ext import build_ext
+
+
+class ZigSDistExt(sdist):
+    def run(self):
+        shutil.copytree("../../src", "src")
```
cmcuza, line 26: Consider using `os.path.join` instead of `../../` although, I tried and it works well in Windows as well.

bindings/python/tersets/__init__.py
```diff
@@ -86,10 +90,12 @@ def compress(values: List[float], method: Method, error_bound: float) -> bytes:
 
     compressed_values = __CompressedValues()
 
-    if (type(method) != Method):
```
cmcuza, line 89: Can you use this PR to update the Enum Method with all the recent algorithms in TerseTS?

cmcuza (changes requested):
I tried the code in Windows and Linux and everything worked well when using `python -m build` followed by `pip install dist/TerseTS-*.whl`. However, when trying to install directly from the code repository using `pip install .` I received an error due to the way the `src/capi.zig` directory is set up. Fixing that in the `setup.py` causes that the build process does not work. I guess you are aware of that, but I think we should somehow bring that up in the README.md inside the python binding. Btw, I found a minor mistake in the example code in `tersets/README.md` . The line:
`compressed_values = compress(uncompressed, method, error_bound)` 
should be replaced by: 
`compressed_values = compress(uncompressed_values, method, error_bound)`
Thanks!

bindings/python/pyproject.toml
skejserjensen: Fixed.

bindings/python/setup.py
skejserjensen: Yes unfortunately the Python tools are really bad at handling any in a parent directory. As I wrote in #52 `setuptools` simply copy from the paths in the manifest to `build/path in the manifest`. So writing `../../src/capi.zig` in the manifest makes `setuptools` copy the files to `bindings/python/build/../../src/capi.zig`which is the same as `bindings` and as a result, the Zig files do not get included in the package. This is why I extended `sdist` and `build_ext`, however, I forgot that it is possible just to run `python -m pip install .` to install a package so I did not extend `install`. Unfortuantely, I have not found a full list of commands supported by `setuptools` but there is one for `distutils` [here](https://docs.python.org/3.11/distutils/apiref.html#distutils.cmd.Command) despite it being deprecated. Maybe it is possible to extend `setuptools` command in a way so that `build_ext` and `install` calls the same code instead of us having to extend each command can build and/or install the package. However, this seems like a larger undertaking so for now I have simple added installation instructions to the readme.

bindings/python/setup.py
skejserjensen: I have kept the `if` statement as is to keep it consistent with the `if` statement on Line 56. Also, I did not wrap `os.remove()`in an exception handler as a compilation error has probably occurred if the `.o` files do not exist.

bindings/python/setup.py
skejserjensen: I used `pathlib` as its path types provide a `parent` variable and that seemed more readable than the functions on `os` that operate on `str`.

bindings/python/tersets/__init__.py
skejserjensen: I am not sure what should be changed, so it is probably safer to make an issue and then update the enum separately.

bindings/python/setup.py
cmcuza: Ok!

bindings/python/tersets/__init__.py
cmcuza: Ok!

cmcuza (approved):

Issue #62, opened 2025-01-08
Automate Publishing to PyPI Using GitHub Actions


cmcuza:
We want a workflow that builds and publishes TerseTS to PyPI whenever we tag a release to avoid manual `twine upload`. For it, we need to add a `.github/workflows/publish.yml` that installs build and twine, compiles the wheel with python -m build, and then uploads to PyPI using `pypa/gh-action-pypi-publish@release/v1`. This ensures anyone can do `pip install tersets` after each release is tagged.

skejserjensen:
Instead of building `wheel`s in GitHub Actions and then publishing them, I suggest we only publish the `sdist` version (Python and Zig source code). This is because our dependency on the `ziglang` package gives us access to Zig when the TerseTS package is installed. Thus, the compilation of the Zig code will happen automatically without requiring that the user installs any additional dependencies manually. By only distributing `sdist` we only have to distribute on archive for each version of the package instead of one per triple (architecture, operating system, etc). In addition, by compiling the code directly on the users machine, Zig can optimize the library specifically for that machine. If we compile on GitHub Actions we have to limit the set of CPU instructions used to a set that is supported by the CPUs users are expected to have with `-mcpu` instead of just compiling with `-mcpu native` on the users machine.

PR #63, opened 2025-01-11
Add support for python -m pip install

skejserjensen:
This PR finishes #52 by resolving the issues with `python -m pip install .` mentioned in the reviews of #61 in preparation for #62. The core issue solved by this PR is to copy `src` to the correct place and delete it afterwards no matter which order the `setuptools` commands are run in. This problem occurs because `python -m build` runs the `setuptools` commands `egg_info`, `sdist`, `egg_info`, `check`, `egg_info`, `bdist_wheel`, `build`, `build_py`, `egg_info`, and `build_ext` if `src` does not exist while  `python -m pip install .` runs the `setuptools` commands `bdist_wheel`, `build`, `build_py`, `egg_info`, and `build_ext` if `src` does not exist. Thus, both `sdist` and `bdist_wheel` have been extended to copy the `src` directory and then delete it again. In addition, checks have been added so `bdist_wheel` only copies the `src` directory if it is run in the repository and `src` does not already exist. As for #61, all methods for package installation should be tested on as many architectures and operating systems as possible.

### DIFF
--- a/bindings/python/README.md
+++ b/bindings/python/README.md
@@ -7,7 +7,7 @@ The TerseTS Python package and native Zig library can be installed from source:
 2. Install the Python `setuptools` package:
    - `python -m pip install setuptools`
 3. Install the TerseTS Python package and native Zig library:
-   - `python -m install .`
+   - `python -m pip install .`
 
 ## Distribution
 The TerseTS Python package and native Zig library can be build from source:
--- a/bindings/python/README.md
+++ b/bindings/python/README.md
@@ -2,11 +2,17 @@
 TerseTS is a library that provides native implementations of methods for lossless and lossy compressing time series in Zig. This package provides a Python interface for TerseTS. For more information about TerseTS, see the repositories main [README file](https://github.com/cmcuza/TerseTS/blob/main/README.md).
 
 ## Installation
-The TerseTS Python package and native Zig library can be build and install from source:
+The TerseTS Python package and native Zig library can be installed from source:
+1. Install the latest version of [Python](https://www.python.org/)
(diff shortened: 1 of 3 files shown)


bindings/python/README.md
```diff
@@ -2,11 +2,17 @@
 TerseTS is a library that provides native implementations of methods for lossless and lossy compressing time series in Zig. This package provides a Python interface for TerseTS. For more information about TerseTS, see the repositories main [README file](https://github.com/cmcuza/TerseTS/blob/main/README.md).
 
 ## Installation
-The TerseTS Python package and native Zig library can be build and install from source:
+The TerseTS Python package and native Zig library can be installed from source:
+1. Install the latest version of [Python](https://www.python.org/)
+2. Install the Python `setuptools` package:
+   - `python -m pip install setuptools`
+3. Install the TerseTS Python package and native Zig library:
+   - `python -m install .`
```
cmcuza, line 10: The right way is `python -m pip install .` or simply `pip install .` 
Maybe we can put both.

cmcuza (changes requested):
All good to go after fixing the commend in the README.md

bindings/python/README.md
skejserjensen: Fixed by writing `python -m pip install .`. I did not add  `pip install .` as calling `pip` directly has been [discouraged for years](https://www.reddit.com/r/learnpython/comments/95w1q3/why_is_it_not_recommended_to_trigger_pip_through/) as far as I am aware.

cmcuza (approved):

PR #68, opened 2025-04-08
Fix issues caused by new Zig release

skejserjensen:
This PR fixes two issues caused by Zig release 0.14.0. First, [Zig 0.14.0](https://ziglang.org/download/0.14.0/release-notes.html#Remove-Anonymous-Struct-Types-Unify-Tuples) changes the type of "untyped anonymous struct literals" such as `const x = .{ .a = 123 }` from "anonymous struct type" to "normal struct type". Second, [Zig 0.14.0](https://ziglang.org/download/0.14.0/release-notes.html#List-of-Deprecations) `std.rand` has been renamed to `std.Random`.

### DIFF
--- a/.github/workflows/zig-build-lint-and-test-on-pr-and-push.yml
+++ b/.github/workflows/zig-build-lint-and-test-on-pr-and-push.yml
@@ -35,7 +35,7 @@ jobs:
     - uses: actions/checkout@v4
     - uses: goto-bus-stop/setup-zig@v2
       with:
-        version: 0.12.0  # Use a stable Zig version
+        version: 0.14.0  # Use a stable Zig version
 
     - name: Zig Build Debug
       run: zig build -Doptimize=Debug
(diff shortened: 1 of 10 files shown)


skejserjensen:
> I see some changes in the syntax like const end = subsequenceStack.pop().?; I will need to check that in the release note. 

The `.?` Optional Unwrap syntax was added to [Zig](https://ziglang.org/documentation/0.3.0/#Table-of-Operators) in version [0.3.0](https://ziglang.org/download/0.3.0/release-notes.html) which was released on [Sep 28, 2018](https://github.com/ziglang/zig/commit/e242f6a609eb4d8367e2bf22ce82e210df3142bb); so this syntax is not new. I can highly recommend reading the entire [Zig Language Reference for the most recent stable release ](https://ziglang.org/documentation/0.14.0/) thoroughly and then read the [release notes](https://ziglang.org/download/0.14.0/release-notes.html) whenever a new version of Zig is released.

> I also tested it locally and everything seems working well. Great job!

Ok, I will merge these changes.

PR #69, opened 2025-04-13
Add linear pointwise compression

sabinabakh:


src/functional/linear_pointwise_compression.zig
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/// Compress using optimal convex hull based Linear Pointwise Approximation.
+pub fn compress(
+    uncompressed_values: []const f64,
+    compressed_values: *ArrayList(u8),
+    error_bound: f32,
+) Error!void {
+    if (uncompressed_values.len < 2) return Error.IncorrectInput;
+    if (error_bound <= 0.0) return Error.UnsupportedErrorBound;
```
cmcuza, line 32: Why is an error bound equal 0 not supported?

src/functional/linear_pointwise_compression.zig
```diff
@@ -0,0 +1,160 @@
+// //! Implementation of Linear Pointwise Approximation (LPA) under the L-infinity norm
```
cmcuza, line 1: Can you also replicate the license information used in the other files here?

src/functional/linear_pointwise_compression.zig
```diff
@@ -0,0 +1,160 @@
...
+    if (error_bound <= 0.0) return Error.UnsupportedErrorBound;
+
+    const epsilon: f64 = @floatCast(error_bound);
+    const allocator = compressed_values.allocator;
+    var convex_hull = try ConvexHull.init(allocator);
+    defer convex_hull.deinit();
+
+    var start: usize = 0;
+
+    while (start < uncompressed_values.len) {
```
cmcuza, line 41: Overall, it would be nice to have more comments on the method's inner workings. For that, you can use the paper as a reference. For example, add a reference to the paper's algorithm that is being implemented. A general rule that we would like to enforce is that, unless trivial, it would be nice to have a comment per line.

src/functional/linear_pointwise_compression.zig
```diff
@@ -0,0 +1,160 @@
...
+
+        const p0 = start;
+        const p1 = last_valid_end;
+        const p0_val = uncompressed_values[p0];
+        const p1_val = uncompressed_values[p1];
+
+        const slope = (p1_val - p0_val) / @as(f64, @floatFromInt(p1 - p0));
+        const intercept = p0_val - slope * @as(f64, @floatFromInt(p0));
+
+        try appendValue(usize, p0, compressed_values);
```
cmcuza, line 71: Why do we need the starting point? In TerseTS we assume a regular sampling of the time starting from 0, so the end point of the previous segment is the starting point of the next segment, so we only need to store one point, right?

src/functional/linear_pointwise_compression.zig
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        // If no valid endpoint was found beyond the start,
+        // linear function cannot be defined, return an error for invalid segment
+        if (last_valid_end == start) return Error.IncorrectInput;
+
+        const p0 = start;
+        const p1 = last_valid_end;
+        const p0_val = uncompressed_values[p0];
+        const p1_val = uncompressed_values[p1];
```
cmcuza, line 66: This algorithm talks about finding A, B, and C points in the convex hull. However, I don't see that logic being implemented here. Can you double-check that the logic of the algorithm is followed closely?

src/functional/linear_pointwise_compression.zig
```diff
@@ -0,0 +1,160 @@
...
+        start = p1 + 1;
+    }
+}
+
+/// Decompress the LPA-compressed stream.
+pub fn decompress(
+    compressed_values: []const u8,
+    decompressed_values: *ArrayList(f64),
+) Error!void {
+    if (compressed_values.len % 32 != 0) return Error.IncorrectInput;
```
cmcuza, line 85: I think this can fixed if only three points are stored as suggested.

src/tersets.zig
```diff
@@ -114,6 +116,13 @@ pub fn compress(
                 error_bound,
             );
         },
+        .LinearPointwiseApproximation => {
```
cmcuza, line 119: How is the method called in the article? Maybe we can use that name to refer to the method instead of calling it LinearPointwiseApproximation because that is a very generic name, and many other methods do the same. If no name is given, maybe we can think of a name based on the logic of the method to avoid generic names.

cmcuza (changes requested):
Thanks a lot for the PR. The comments below seek to improve the quality of the code base before merging into `main`. Besides the comments below, please add a more informative comment in the PR itself. There, explain the PR in more detail regarding the files modified and the algorithm implemented.

src/functional/linear_pointwise_compression.zig
sabinabakh: I’ve replicated the licence as instructed

src/functional/linear_pointwise_compression.zig
sabinabakh: Only the endpoint is saved in the updated version

src/functional/linear_pointwise_compression.zig
sabinabakh: I’ve adjusted the method and included more comments

src/tersets.zig
sabinabakh: I’ve renamed the method to OptimalPiecewiseLinearApproximation which is very close to how it’s defined in the paper

src/functional/abc_linear_compression.zig
```diff
@@ -0,0 +1,308 @@
...
+
+        // Find v(l_i): the vertex that maximizes deviation from the side l_i
+        var max_dev: f64 = -1.0;
+        var pivot_idx: usize = i;
+
+        for (i + 1..len) |j| {
+            const C = convex_hull.at(j);
+
+            // Deviation of point C from line l_i
+            const pred_value = side_slope * (@as(f64, @floatFromInt(C.time)) - @as(f64, @floatFromInt(A.time))) + A.value;
```
cmcuza, line 74: I think doing this looks better.

`const delta_time:f64 = @floatFromInt(C.time - A.time)
const pred_value = side_slope * delta_time + A.value;
`

src/functional/abc_linear_compression.zig
```diff
@@ -0,0 +1,308 @@
...
+    var finished = false;
+
+    while (!finished) {
+        if (i + 1 >= len) break; // No more sides
+
+        const A = convex_hull.at(i);
+        const B = convex_hull.at(i + 1);
+
+        // Compute slope of side l_i
+        const side_slope = (B.value - A.value) / (@as(f64, @floatFromInt(B.time)) - @as(f64, @floatFromInt(A.time)));
```
cmcuza, line 64: I think doing this looks better.

`
const delta_time = @floatFromInt(B.time-A.time);
const side_slope = (B.value - A.value) / delta_time;`

src/functional/abc_linear_compression.zig
```diff
@@ -0,0 +1,308 @@
...
+            // x-internal -> optimal segment found
+            finished = true;
+        }
+    }
+
+    // Once finished, line from A to B is the optimal approximation
+    const start = convex_hull.at(i);
+    const end = convex_hull.at(i + 1);
+
+    const slope = (end.value - start.value) / (@as(f64, @floatFromInt(end.time)) - @as(f64, @floatFromInt(start.time)));
```
cmcuza, line 109: Something similar can be done here with the delta time as suggested before.

cmcuza (changes requested):
Overall, the PR is much better, but I think there is a bug in the `findABCOptimalSegment` function. I left a comment to reproduce the error. It would be nice to have that fixed before merging.

src/functional/abc_linear_compression.zig
sabinabakh: I've refactored the formula accordingly

src/functional/abc_linear_compression.zig
sabinabakh: Completed.

src/functional/abc_linear_compression.zig
sabinabakh: Completed.

src/functional/abc_linear_compression.zig
```diff
@@ -0,0 +1,308 @@
...
+pub fn compress(
+    uncompressed_values: []const f64,
+    compressed_values: *ArrayList(u8),
+    allocator: mem.Allocator,
+    error_bound: f32,
+) Error!void {
+    if (uncompressed_values.len < 2) return Error.IncorrectInput;
+    if (error_bound <= 0.0) return Error.UnsupportedErrorBound;
+
+    const epsilon: f64 = @floatCast(error_bound);
```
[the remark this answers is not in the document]
sabinabakh: You are right, this is redundant. This has now been changed: error_bound stays f32.

src/functional/abc_linear_compression.zig
```diff
@@ -0,0 +1,308 @@
...
+const Method = tersets.Method;
+
+const shared = @import("../utilities/shared_structs.zig");
+const DiscretePoint = shared.DiscretePoint;
+const LinearFunction = shared.LinearFunction;
+const ConvexHull = @import("../utilities/convex_hull.zig").ConvexHull;
+const Segment = shared.Segment;
+
+/// Helper to serialize a value into bytes
+fn appendValue(comptime T: type, value: T, compressed: *ArrayList(u8)) !void {
```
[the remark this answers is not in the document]
sabinabakh: Moved compress and decompress methods to the top of the file.

src/functional/abc_linear_compression.zig
```diff
@@ -0,0 +1,308 @@
...
+        const noise = if (i % 2 == 0) @as(f64, 0.05) else @as(f64, -0.05);
+        try uncompressed_values.append(3.0 * @as(f64, @floatFromInt(i)) + 4.0 + noise);
+    }
+
+    try tester.testCompressAndDecompress(uncompressed_values.items, allocator, Method.ABCLinearApproximation, error_bound, tersets.isWithinErrorBound);
+}
+
+test "random lines and error bound compress and decompress" {
+    const allocator = std.testing.allocator;
+    const error_bound: f32 = 0.01;
```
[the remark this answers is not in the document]
sabinabakh: This issue is now resolved. The current implementation keeps track of visited point candidates for A and B to prevent an infinite loop.

src/functional/abc_linear_compression.zig
```diff
@@ -0,0 +1,308 @@
...
+            const y = slope * x + intercept;
+            try decompressed_values.append(y);
+        }
+
+        start = end + 1;
+        i += 3;
+    }
+}
+
+test "compresses and decompresses perfect linear signal" {
```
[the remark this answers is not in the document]
sabinabakh: After adding some general fixes, the issue resolved itself. I was unable to reproduce this exact scenario

cmcuza:
Hi, almost there but, still some things to fix. I think a proper merge with the main branch must be done. Right now this is the error:

src/functional/abc_linear_compression.zig:51:50: error: no error named 'IncorrectInput' in 'error{OutOfMemory,UnknownMethod,UnsupportedInput,UnsupportedErrorBound,ItemNotFound,EmptyConvexHull}'
    if (uncompressed_values.len < 2) return Error.IncorrectInput;
                                            ~~~~~^~~~~~~~~~~~~~~
src/functional/abc_linear_compression.zig:126:54: error: no error named 'IncorrectInput' in 'error{OutOfMemory,UnknownMethod,UnsupportedInput,UnsupportedErrorBound,ItemNotFound,EmptyConvexHull}'
    if (compressed_values.len % 24 != 0) return Error.IncorrectInput;

cmcuza (changes requested):

cmcuza (approved):

PR #71, opened 2025-05-07
Add Visvalingam-Whyatt compression algorithm

cmcuza:
This PR adds Visvalingam-Whyatt (VW) [1] line simplification algorithm. For this PR we are adding a new category of compression method based on line simplification. Specifically, VW reduces points based on the area under the curve created by triples of points in the time series. The PR includes the testing of the method. 

[1] https://doi.org/10.1179/000870493786962263

Issue #75, opened 2025-05-12
Add better support for `nan` and `non-finite` f64


cmcuza:
Yes, having support for `NaN` does seem to make many calculations and checks much more complicated, but since it is a value supported by `f64`, we have to either support it or check that it does not exist in a `f64` slice. I agree that the task of adding the helper methods and adding the correct checks to the start of each compression method should probably be done in a separate PR as a response to a separate issue. Please create the easy and link back to this discussion so we keep all of the information. Afterwards, feel free to resolve the discussion.

_Originally posted by @skejserjensen in https://github.com/cmcuza/TerseTS/pull/74#discussion_r2084887225_

skejserjensen:
No only should `nan`, `+inf`, and `-inf` be supported, but EVERY SINGLE combination of `f64` values should either be compressed/decompressed correctly by EACH method or reject the inputs they do not support. It should be impossible for users to lose data when using TerseTS.

PR #76, opened 2025-05-12
Add bottom up line simplification

cmcuza:
This PR adds the implementation of the "Bottom-Up" algorithm as described in the paper [1]. I have been merging with the other two branches created from the previous open PRs to stay up to date, but I think it was a bad idea. Now you will see that this PR is also adding the VW algorithm, which is actually #71. I think once merged, that should not be a problem. In any case, the main changes are in the `src/line_simplification/bottom_up.zig` file. So you can ignore the changes in the rest of the files until accepting #71. 

[1] https://doi.org/10.1109/ICDM.2001.989531


src/line_simplification/bottom_up.zig
```diff
@@ -0,0 +1,519 @@
...
+/// is used to allocate memory for the HashedPriorityQueue used in the implementation.
+/// If an error occurs it is returned.
+pub fn compress(
+    uncompressed_values: []const f64,
+    compressed_values: *ArrayList(u8),
+    allocator: mem.Allocator,
+    error_bound: f32,
+) Error!void {
+
+    // If we have 2 or fewer points, store them without compression.
```
skejserjensen, line 59: , we store

src/line_simplification/bottom_up.zig
```diff
@@ -0,0 +1,519 @@
...
+    if (uncompressed_values.len <= 2) {
+        try appendValue(f64, uncompressed_values[0], compressed_values);
+        try appendValue(usize, 1, compressed_values);
+        try appendValue(f64, uncompressed_values[1], compressed_values);
+        return;
+    }
+
+    if (error_bound < 0) {
```
skejserjensen, line 67: Maybe check error bound first so the error is also returned for inputs of zero, one, or two values.

src/line_simplification/bottom_up.zig
```diff
@@ -0,0 +1,519 @@
...
+        void,
+        compareSegmentMergeCost,
+        SegmentMergeCostHashContext,
+    ).init(allocator, {});
+    defer heap.deinit();
+
+    // Compute all pairwise merging cost of the segments.
+    try buildInitialPairwiseSegmentCost(&heap, uncompressed_values);
+
+    // Placeholder for the segment cost to be used in the loop to search for neighboring segments.
```
skejserjensen, line 84: Maybe explain the concept of left and right segments here as it is used throughout the `while` loop below.

src/line_simplification/bottom_up.zig
```diff
@@ -0,0 +1,519 @@
...
+            // Update the cost of the right segment with the computed merge cost.
+            right_seg.cost = merge_cost;
+        }
+
+        // Update the right segment in the heap.
+        try heap.update(right_seg, right_seg);
+    }
+
+    // Sort remaining points by original index to preserve order.
+    std.mem.sort(SegmentMergeCost, heap.items[0..heap.len], {}, SegmentMergeCost.firstThan);
```
skejserjensen, line 159: Maybe use `mem` and `math` directly throughout the file as they are imported and assigned to a `const` at the start of the file.

src/line_simplification/bottom_up.zig
```diff
@@ -0,0 +1,519 @@
...
+        const seg_end = heap.items[index].seg_end;
+        // Check if the segment has more than two points.
+        if (seg_start + 1 < seg_end) {
+            // If the segment has only two points, directly append the start value,
+            // the end index, and the end value to the compressed representation.
+            try appendValue(f64, uncompressed_values[seg_start], compressed_values);
+            try appendValue(usize, seg_end, compressed_values);
+            try appendValue(f64, uncompressed_values[seg_end], compressed_values);
+        } else {
+            // If the segment has many points, compute the linear regression line
```
skejserjensen, line 173: has more than two points

src/line_simplification/bottom_up.zig
```diff
@@ -0,0 +1,519 @@
...
+
+    // Sort remaining points by original index to preserve order.
+    std.mem.sort(SegmentMergeCost, heap.items[0..heap.len], {}, SegmentMergeCost.firstThan);
+
+    // Output compressed series: (seg_start, index, end_value) pairs.
+    for (0..heap.len) |index| {
+        const seg_start = heap.items[index].seg_start;
+        const seg_end = heap.items[index].seg_end;
+        // Check if the segment has more than two points.
+        if (seg_start + 1 < seg_end) {
```
skejserjensen, line 166: Can a segment have a single value and is a separate branch required to handle this case?

src/line_simplification/bottom_up.zig
```diff
@@ -0,0 +1,519 @@
...
+
+            // Interpolate the values between the start and end points of the current segment.
+            while (current_timestamp < end_point.time) : (current_timestamp += 1) {
+                const y: f64 = slope * @as(f64, @floatFromInt(current_timestamp)) + intercept;
+                try decompressed_values.append(y);
+            }
+            try decompressed_values.append(end_point.value);
+            first_timestamp = current_timestamp + 1;
+        } else {
+            // If the start and end points are the distance 1,
```
skejserjensen, line 229: Unclear what "are the distance 1" means and why the comment is formatted as two very short equal-length lines.

src/line_simplification/bottom_up.zig
```diff
@@ -0,0 +1,519 @@
...
+            // append the start point and end points directly.
+            try decompressed_values.append(start_point.value);
+            try decompressed_values.append(end_point.value);
+            first_timestamp += 2;
+        }
+    }
+}
+
+/// `SegmentMergeCost` represents a segment and its associated `cost` with the next segment,
+/// which is calculated based RSSE. The `cost` is used to determine the order in which segments
```
skejserjensen, line 239: "based on RSSE" or "using RSSE"?

src/line_simplification/bottom_up.zig
```diff
@@ -0,0 +1,519 @@
...
+        // Compute the merge cost of merging the previous and current segments.
+        const merge_cost = mergeCost(uncompressed_values, previous_seg, current_seg);
+        previous_seg.cost = merge_cost;
+        try heap.add(previous_seg);
+
+        seg_id += 1;
+        previous_seg = current_seg;
+    }
+
+    // Insert the previous segment which is probably the last one and has infinite cost.
```
skejserjensen, line 334: Explain in more detail why the previous segment is probably the last and why it is important.

src/line_simplification/bottom_up.zig
```diff
@@ -0,0 +1,519 @@
...
+
+/// Incremental cost of merging the neighbouring segments `seg_one` and `seg_two`.
+fn mergeCost(uncompressed_values: []const f64, seg_one: SegmentMergeCost, seg_two: SegmentMergeCost) f64 {
+    const merged_start = @min(seg_one.seg_start, seg_two.seg_start);
+    const merged_end = @max(seg_one.seg_end, seg_two.seg_end);
+    return computeRMSE(uncompressed_values, merged_start, merged_end);
+}
+
+/// Compute the linear regression that minimizes the Root-Mean-Squared-Errors (RMSE).
+fn computeLinearRegression(values: []const f64, seg_start: usize, seg_end: usize) LinearFunction {
```
skejserjensen, line 361: Move the functions used by the algorithms added by #76 and #77, such as `computeLinearRegression()`, to a shared module and use throughout TerseTS instead of duplicating them. See #77 for additional comment about the duplicate functions.

src/line_simplification/bottom_up.zig
```diff
@@ -0,0 +1,519 @@
...
+    // Output buffer.
+    var uncompressed_values = ArrayList(f64).init(allocator);
+    defer uncompressed_values.deinit();
+    var compressed_values = ArrayList(u8).init(allocator);
+    defer compressed_values.deinit();
+    var decompressed_values = ArrayList(f64).init(allocator);
+    defer decompressed_values.deinit();
+    const error_bound: f32 = 0.0;
+
+    // Bottom Up cannot handle very large value due to numerical issues with `math.order()`.
```
skejserjensen, line 456: See this [comment](https://github.com/cmcuza/TerseTS/pull/77/files#r2094155722), simply making the tests pass does not make the code correct.

skejserjensen (changes requested):

src/line_simplification/bottom_up.zig
cmcuza: Fixed.

src/line_simplification/bottom_up.zig
cmcuza: Fixed.

src/line_simplification/bottom_up.zig
cmcuza: Fixed.

src/line_simplification/bottom_up.zig
cmcuza: Fixed all around.

src/line_simplification/bottom_up.zig
cmcuza: Fixed.

src/line_simplification/bottom_up.zig
cmcuza: Fixed.

src/line_simplification/bottom_up.zig
cmcuza: fixed.

src/line_simplification/bottom_up.zig
cmcuza: Fixed.

src/line_simplification/bottom_up.zig
cmcuza: I already had plans to do this as an independent PR after completing this one.

src/line_simplification/bottom_up.zig
cmcuza: I know, I think this needs to be solved as described in #75. In this particular case, very extreme floating points raises an error in the `math.order()` function. I think that as part of the preprocessing step described in #75, we can also detect very big extreme floating points and avoid calling the function with an error code of `UnsuportedInput` for this function.

src/line_simplification/bottom_up.zig
cmcuza: I consider that case on the sliding window #77, but for some reason forgot to consider it here. Now it is fixed.

cmcuza:
Besides the requested changes, I also changed the logic of the computeRMSE function. There was a conceptual mistake since I put the method as part of a line simplification, meaning that all points should be part of the original time series (kind of a sampling method). Before I was encoding the line with min RMSE.

Issue #55, opened 2024-11-17
Update TerseTS to support various compression parameterizations


cmcuza:
This issue is to enable TerseTS to accept different kinds of parameters to the compression algorithms. For example, PWCH does not receive an error bound, instead, it receives a minimum compression ratio (which internally is treated as the number of bins in the histogram). Likewise, other methods will require their parameterization instead of the current generic `error_bound`.

_Originally posted by @skejserjensen in https://github.com/cmcuza/TerseTS/pull/50#discussion_r1845217883_

skejserjensen:
For the full discussion, see this [conversation in #50](https://github.com/cmcuza/TerseTS/pull/50#discussion_r1827039060).

skejserjensen:
As specified in #46 the API of TerseTS is already inconsistent as some languages use a `Configuration` `struct` while other use multiple parameters for `method` and `error_bound`.

So the first question to answer is which API to standardize on. I propose using a `Configuration` `struct` everywhere as it can then differ per method instead of having parameters that must all be passed arguments despite each method only using a subset, e.g., `error_bound` is used by many but not all compression methods and `number_of_buckets` is only used by a few compression methods.

The next question is how this `Configuration` `struct` should look. One option could be to simply have have the total set of supported parameters in one `struct`, i.e., have the `struct` contain `method`, `error_bound`, `number_of_buckets`, etc. This is very simple and allows any combination of parameters to be passed to any compression method which can then just check that all the parameters it needs is present and maybe that nothing unsupported is set. To make it simpler for the user to create this `struct`, appropriate APIs for each language should be created, e.g., an anonymous `struct` could maybe be used in [Zig](https://www.reddit.com/r/Zig/comments/sanpzf/zig_best_practices_emerging_patterns/?rdt=33439), optional arguments in Python, and a builder pattern in [Rust](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html).

[...]

PR #79, opened 2025-05-16
Add support for compression parametrization

cmcuza:
This PR addresses the issue described in #55. This PR introduces a unified and extensible compression API across Zig, C, and Python. Following the discussion in https://github.com/cmcuza/TerseTS/issues/55#issuecomment-2500030000, a new `Configuration` struct is used to encapsulate the compression method and its associated parameters. This includes support for method-specific parameter structs (e.g., `FunctionalParams`, `HistogramParams`, `LineSimplificationParams`) as well as a simplified fallback `BasicParams` for users who only want to specify an error bound. The new API automatically maps `BasicParams` to the appropriate method-specific configuration. All these new configurations are part of a new file called `params.zig`.
In addition to changes in Zig to adapt to these changes, the C interface has been updated to reflect this design with enums and structs. Python bindings now support both simple and advanced configurations, and the `compress()` function dynamically determines how to construct and pass the correct parameter struct. I think it looks good, but I know it will generate a lot of work and a few iterations before merging.

### DIFF
--- a/bindings/c/tersets.h
+++ b/bindings/c/tersets.h
@@ -1,32 +1,124 @@
+// Copyright 2024 TerseTS Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef TERSETS_H
+#define TERSETS_H
+
 #include <stdint.h>
+#include <stddef.h> // for size_t
 
-// A pointer to uncompressed values and the number of values.
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// -----------------------------------------------------------------------------
+// Data Structures
+// -----------------------------------------------------------------------------
+
+/// A pointer to uncompressed values and the number of values.
 struct UncompressedValues {
-  double const * const data;
-  uintptr_t const len;
+    const double* data;
+    uintptr_t len;
 };
 
-// A pointer to compressed values and the number of bytes.
+/// A pointer to compressed values and the number of bytes.
 struct CompressedValues {
-  uint8_t const * const data;
-  uintptr_t const len;
+    const uint8_t* data;
+    uintptr_t len;
 };
 
-// Configuration to use for compression and/or decompression.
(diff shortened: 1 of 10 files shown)


cmcuza:
Ok, not quite ready. The test is failing on ReleaseFast mode. Weird! I did not test this locally before committing. Not sure why that.

cmcuza:
Great, now all tests pass. It is time to require a revision @skejserjensen.
The PR extended itself over almost all files. I also ended up creating a PMC-Midrange using a relative error bound to test that the parametrization works well. Still, some things can and will be improved. For example, we can now create proper validation of the type of parametrization. Regardless, this was an important feature that needed to be included, and I think the current solution gets us close to it.

skejserjensen (changes requested):
While I agree that having a setup with static checking of the parameters names and types is nice, I think the current version adds too much complexity for too little gain. Maybe we can get some inspiration from `libpressio` ([see](https://github.com/szcompressor/fz_tutorial/blob/master/exercises/1_basics/python/basics.py)) and use a standard textual format so each compression methods can receive exactly the arguments it needs as a string. JSON could be an option as [Zig's standard library](https://ziglang.org/documentation/master/std/) includes a [JSON module](https://ziglang.org/documentation/master/std/#std.json). There is also Zig's own [Zon format](https://ziglang.org/documentation/master/std/#std.zon), but it may be nice to use a more standard format so it can be easily generated  by other tools. If we chose a textual format for parameters, I think the most important part is to no require users to install dependencies and the format should also be easy for the user to write.

skejserjensen:
The following is an example of how using JSON for the configuration could work:

```zig
const std = @import("std");
const json = std.json;

var gpa = std.heap.GeneralPurposeAllocator(.{}){};
const allocator = gpa.allocator();

// A Configuration struct is defined for each compression method which
// specifies exactly which parameters the compression methods requires.
const Configuration = struct { relative_error_bound: f32, bucket_size: usize };

pub fn main() !void {
    // Code executed at the start of each compression method.
    const parsed = try json.parseFromSlice(
        Configuration,
        allocator,
        \\{ "relative_error_bound": 5.0, "bucket_size": 25 }
    ,
        .{},
    );
    defer parsed.deinit();

    std.debug.print("{}", .{parsed});
}
```
Each compression method defines its own `Configuration` struct in the same file as the compression method. The `compress()` function in `tersets.zig` is changed to something like `compress(uncompressed_values: []const f64, allocator: Allocator, method: Method, configuration: []u8)`. This allows `compress()` to select the compression method to use based on `method` and then call that compression method with `uncompressed_values`, `allocator`, and `configuration`. The function that implements the compression method can then do as shown in the `main()` above. Since `json.parseFromSlice()` checks that the JSON is correct, only contains the fields in `Configuration`, and all of these fields have the correct types, no additional checks are needed.

skejserjensen:
Is the branch not deleted on purpose?

PR #83, opened 2025-06-26
Improve SimPiece testing

cmcuza:
This PR improves SimPiece testing by generating random error bounds and multiple random linear functions in the uncompressed data. It also adapts the testing to the new random generation framework.


skejserjensen (approved):
Approved as it improves the testing code, but #75 still needs to be fixed for all compression methods.

PR #86, opened 2025-06-30
Add Run Length Encoding

cmcuza:
This PR adds the Run Length Encoding (RLE) algorithm. This algorithm belongs to the lossless encoding methods that form part of TerseTS. Users can use RLE in combination with other compression methods, potentially after quantization. I also added the necessary testing.

### DIFF
--- a/src/lossless_encoding/run_length_encoding.zig
+++ b/src/lossless_encoding/run_length_encoding.zig
@@ -79,7 +79,7 @@ pub fn decompress(compressed_values: []const u8, decompressed_values: *ArrayList
 
 /// Append `value` of `type` determined at compile time to `compressed_values`.
 fn appendValue(comptime T: type, value: T, compressed_values: *std.ArrayList(u8)) !void {
-    // Compile-time type check
+    // Compile-time type check.
     switch (@TypeOf(value)) {
         f64, usize => {
             const value_as_bytes: [8]u8 = @bitCast(value);
--- a/src/lossless_encoding/run_length_encoding.zig
+++ b/src/lossless_encoding/run_length_encoding.zig
@@ -1,0 +1,136 @@
+// Copyright 2024 TerseTS Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Implementation of "Run-Length-Encoding" for compressing and decompressing time series data.
+//! This implementation compresses sequences of repeated values into a single value and a count,
+//! allowing for efficient storage and transmission of data with many repeated values.
+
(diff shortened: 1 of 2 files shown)


src/lossless_encoding/run_length_encoding.zig
```diff
@@ -0,0 +1,136 @@
...
+    );
+}
+
+test "run length encoding compresses repeated values" {
+    const allocator = testing.allocator;
+
+    var uncompressed_values = ArrayList(f64).init(allocator);
+    defer uncompressed_values.deinit();
+
+    const distinct_elements: usize = @intFromFloat(@round(tester.generateBoundedRandomValue(
```
skejserjensen, line 109: Unclear what `distinct_elements` are used for, why `@round` is applied, and why 10 to 50 is used instead of any `f64` type.

src/lossless_encoding/run_length_encoding.zig
```diff
@@ -0,0 +1,136 @@
...
+        f64,
+        10,
+        50,
+        undefined,
+    )));
+
+    for (0..distinct_elements) |_| {
+        const random_value = tester.generateBoundedRandomValue(f64, -1e-16, 1e16, undefined);
+        const repeat: usize = @intFromFloat(@round(tester.generateBoundedRandomValue(
```
skejserjensen, line 118: Maybe make `tester.generateBoundRandomInteger()` using whatever functionality Zig includes for generating random integers instead of generating `f64` and then converting to integers.

skejserjensen (changes requested):

src/lossless_encoding/run_length_encoding.zig
skejserjensen: The magic numbers are still unclear. Why 10 to 50? Also, why are integers created when the `compress()` function being tested accept `[]const f64`? Since `[]const f64` is accepted, sequences of any combination of `f64` should be generated and used for testing.

src/lossless_encoding/run_length_encoding.zig
skejserjensen: The code is now simpler to understand due to the helper function, but the magic numbers 10 to 20 are still unclear.

src/lossless_encoding/run_length_encoding.zig
cmcuza: Regarding f64, I added a new helper function to generate any f64 value that can be called during testing. Regarding the magic numbers 10 and 50, it was the simplest way to generate a bounded integer to indicate the number of distinct values to generate during the testing. It could have been any pair of integer numbers, but any of them would trigger this comment anyway. I will keep these two numbers, but add a bit more detailed comment to explain why 10-50 is a good range for testing.

cmcuza: I followed the comment, creating a `global_at_least` and `global_at_most` variables in the `tester`. Maybe in the future these variables can be reused, but for now, only this test function needs them. Upcoming PRs will be adapted to use these variables if needed.

src/lossless_encoding/run_length_encoding.zig
cmcuza: I followed the comment, creating a `global_at_least` and `global_at_most` variables in the `tester`. Maybe in the future these variables can be reused, but for now, only this test function needs them. Upcoming PRs will be adapted to use these variables if needed.

skejserjensen (approved):

cmcuza:
It needs approval again after being merged with the main branch.

skejserjensen (approved):

PR #88, opened 2025-07-01
Add Method enum synchronization test

cmcuza:
This PR ensures that the Python Method enum matches the Zig implementation. Additionally, it adds a C binding Method enum and a test that matches the Zig implementation through a Python unittest. The tests check for correct Method enum ordering and verify that enums stay synchronized. This PR also updates the README's C example to reference the enum Method.

### DIFF
--- a/bindings/c/tersets.h
+++ b/bindings/c/tersets.h
@@ -16,6 +16,7 @@ enum Method {
   BottomUp = 11,
   MixPiece = 12,
   BitPackedQuantization = 13,
+  RunLengthEncoding = 14,
 };
 
 // A pointer to uncompressed values and the number of values.


skejserjensen (changes requested):

skejserjensen:
> > Seems the two comments are still valid.
> 
> Yes, you have not resolved them!

My point was that the code did not seem to have been updated to fix the issues raised in the comments and no response was provided to explain why the code was not changed either.

cmcuza:
> > Interesting, I dont know why you cannot see, but the comments were addressed. I have also added the comments explaining how. Below a screenshot of how the comments are displayed to me.
> 
> The reason I cannot see the comments is that they are `pending` as clearly shown by the badge next to your name. I do not know why GitHub has marked your comments as pending.

Weird, I did not know this feature existed and how it came to be. Now it should work.

PR #99, opened 2025-09-20
Refactor Python's binding

cmcuza:
### Summary

This PR refactors the Python bindings for compress and decompress to provide a single, unified public API with automatic fast paths:
* If NumPy is installed and the user passes an `np.ndarray[float64]`, we pass around pointers to the same memory instead of duplicating it (typically ~2x faster on large arrays).
* If the user passes a `list/tuple` of floats, we use an optimized ctypes path with a contiguous `c_double[]` buffer ( ~50x faster than the previous implementation).

#### Key improvements:
* Return real bytes for `compress` via `ctypes.string_at` (no Python list-of-ints).
* Use unsigned byte pointers (`c_ubyte*`) for compressed buffers to match Zig’s `[]u8`.
* Add explicit ctypes function signatures (argtypes/restype) for `compress`/`decompress` to prevent silent ABI mismatches.
* In decompress, when NumPy is present, we create a **view** of the input bytes and immediately materialize a Python list (`tolist()`), ensuring no dangling pointers to Zig-owned memory.

### Files Added/Modified

- Python binding at `binding/Python/tersets/__init__.py`

### Testing

I significantly tested the new binding using NumPy and list/tuple arrays as input. I do not include them as part of the PR because NumPy is not installed as part of our testing workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,47 +78,46 @@ TerseTS provides `./src/tersets.zig` as the single access point and two main fun
 #include "tersets.h"
 #include <stdio.h>
 
-int main() {
-   double uncompressed_values[] = {1.0, 2.0, 3.0, 4.0, 5.0};
-   struct UncompressedValues uncompressed_values = {data, 5};
+int main(void) {
+    // Input data
+    double data[] = {1.0, 2.0, 3.0, 4.0, 5.0};
+    struct UncompressedValues uncompressed_values = { data, 5 };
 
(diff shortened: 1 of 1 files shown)


skejserjensen (changes requested):

skejserjensen (changes requested):
See this [comment](https://github.com/cmcuza/TerseTS/pull/99#discussion_r2431698765).

skejserjensen (approved):

PR #104, opened 2025-10-30
Restructure TerseTS API

cmcuza:
### Summary

This PR updates the public compression API (Zig + C + Python) to accept JSON-based method configurations instead of a fixed binary configuration struct. This is necessary because different TerseTS methods now require different configuration shapes (e.g. abs_error_bound, aggregate_error_type + aggregate_error_bound, histogram_bins_number, empty config for RLE). Making the API take a JSON string keeps the C/Python boundary simple and lets us evolve method-specific configs without changing the FFI every time.

- Closes #
  This PR closes #79 .

### DIFF
--- a/src/configuration.zig
+++ b/src/configuration.zig
@@ -106,346 +106,298 @@ pub fn parse(
     return parsed_value;
 }
 
-test "check abs error bound configuration" {
+test "parse valid AbsoluteErrorBound" {
     const allocator = std.testing.allocator;
-    const valid_configuration = try parse(
+    const config = try parse(
         allocator,
         AbsoluteErrorBound,
-        \\ { "abs_error_bound": 0.1 }
+        \\{ "abs_error_bound": 0.1 }
         ,
     );
-    try testing.expectEqual(@TypeOf(valid_configuration), AbsoluteErrorBound);
-    try testing.expectEqual(valid_configuration.abs_error_bound, 0.1);
+    try testing.expectEqual(@TypeOf(config), AbsoluteErrorBound);
+    try testing.expectEqual(config.abs_error_bound, 0.1);
+}
 
-    _ = parse(
-        allocator,
-        AbsoluteErrorBound,
-        \\ { "ab_eror_bund": 0.1 }
-        ,
-    ) catch |err| {
-        try testing.expectEqual(error.InvalidConfiguration, err);
-        return;
-    };
-
-    try testing.expectFmt(
-        "",
-        "The configuration is supposed to be invalid",
-        .{},
-    );
-
+test "parse AbsoluteErrorBound with wrong field name" {
+    const allocator = std.testing.allocator;
     _ = parse(
         allocator,
         AbsoluteErrorBound,
-        \\ { "abs_error_bound": 0.1.2 }
+        \\{ "ab_eror_bund": 0.1 }
         ,
     ) catch |err| {
         try testing.expectEqual(error.InvalidConfiguration, err);
         return;
     };
+    unreachable;
+}
 
-    try testing.expectFmt(
-        "",
-        "The configuration is supposed to be invalid",
-        .{},
-    );
-
+test "parse AbsoluteErrorBound with invalid float" {
+    const allocator = std.testing.allocator;
     _ = parse(
         allocator,
         AbsoluteErrorBound,
-        \\ { "abs_error_bound": -0.1 }
+        \\{ "abs_error_bound": 0.1.2 }
         ,
     ) catch |err| {
         try testing.expectEqual(error.InvalidConfiguration, err);
         return;
     };
-
-    try testing.expectFmt(
-        "",
-        "The error bound is supposed to be positive",
-        .{},
-    );
+    unreachable;
 }
 
-test "check rel error bound configuration" {
+test "parse AbsoluteErrorBound with negative value" {
     const allocator = std.testing.allocator;
-    const valid_configuration = try parse(
-        allocator,
-        RelativeErrorBound,
-        \\ { "rel_error_bound": 0.1 }
-        ,
-    );
-    try testing.expectEqual(@TypeOf(valid_configuration), RelativeErrorBound);
-    try testing.expectEqual(valid_configuration.rel_error_bound, 0.1);
-
     _ = parse(
         allocator,
-        RelativeErrorBound,
-        \\ { "are_eror_bund": 0.1 }
+        AbsoluteErrorBound,
+        \\{ "abs_error_bound": -0.1 }
         ,
     ) catch |err| {
         try testing.expectEqual(error.InvalidConfiguration, err);
         return;
     };
+    unreachable;
+}
 
-    try testing.expectFmt(
-        "",
-        "The configuration is supposed to be invalid",
-        .{},
+test "parse valid RelativeErrorBound" {
+    const allocator = std.testing.allocator;
+    const config = try parse(
+        allocator,
+        RelativeErrorBound,
+        \\{ "rel_error_bound": 0.1 }
+        ,
     );
+    try testing.expectEqual(@TypeOf(config), RelativeErrorBound);
+    try testing.expectEqual(config.rel_error_bound, 0.1);
+}
 
+test "parse RelativeErrorBound with wrong field name" {
+    const allocator = std.testing.allocator;
     _ = parse(
         allocator,
         RelativeErrorBound,
-        \\ { "rel_error_bound": 0.1.2 }
+        \\{ "are_eror_bund": 0.1 }
         ,
     ) catch |err| {
         try testing.expectEqual(error.InvalidConfiguration, err);
         return;
     };
+    unreachable;
+}
 
-    try testing.expectFmt(
-        "",
-        "The configuration is supposed to be invalid",
-        .{},
-    );
-
+test "parse RelativeErrorBound with invalid float" {
+    const allocator = std.testing.allocator;
     _ = parse(
         allocator,
         RelativeErrorBound,
-        \\ { "rel_error_bund": -0.1 }
+        \\{ "rel_error_bound": 0.1.2 }
         ,
     ) catch |err| {
         try testing.expectEqual(error.InvalidConfiguration, err);
         return;
     };
-
-    try testing.expectFmt(
-        "",
-        "The error bound is supposed to be positive",
-        .{},
-    );
+    unreachable;
 }
 
-test "check auc error bound configuration" {
+test "parse RelativeErrorBound with negative value" {
     const allocator = std.testing.allocator;
-    const valid_configuration = try parse(
-        allocator,
-        AreaUnderCurveError,
-        \\ { "area_under_curve_error": 0.1 }
-        ,
-    );
-    try testing.expectEqual(@TypeOf(valid_configuration), AreaUnderCurveError);
-    try testing.expectEqual(valid_configuration.area_under_curve_error, 0.1);
-
     _ = parse(
         allocator,
-        AreaUnderCurveError,
-        \\ { "auc_error": 0.1 }
+        RelativeErrorBound,
+        \\{ "rel_error_bund": -0.1 }
         ,
     ) catch |err| {
         try testing.expectEqual(error.InvalidConfiguration, err);
         return;
     };
+    unreachable;
+}
 
-    try testing.expectFmt(
-        "",
-        "The configuration is supposed to be invalid",
-        .{},
+test "parse valid AreaUnderCurveError" {
+    const allocator = std.testing.allocator;
+    const config = try parse(
+        allocator,
+        AreaUnderCurveError,
+        \\{ "area_under_curve_error": 0.1 }
+        ,
     );
+    try testing.expectEqual(@TypeOf(config), AreaUnderCurveError);
+    try testing.expectEqual(config.area_under_curve_error, 0.1);
+}
 
+test "parse AreaUnderCurveError with wrong field name" {
+    const allocator = std.testing.allocator;
     _ = parse(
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -86,15 +87,14 @@ int main(void) {
...
+     const char *cfg = "{\"abs_error_bound\": 0.01}";
 
     // Output buffers for compressed/decompressed data.
     struct CompressedValues   compressed_values = {0};
     struct UncompressedValues decompressed_values = {0};
 
-    // Compression.
-    int32_t result = compress(uncompressed_values, &compressed_values, config);
+    // Compression. The C API mirrors the `Method` enum from TerseTS.
+    int32_t result = compress(uncompressed_values, &compressed_values, SwingFilter, cfg);
```
skejserjensen: Why is the configuration named `cfg` here instead of the more informative `configuration` as in the `Python` example? I see no upside to use `cfg` instead of `configuration` so I suggest using `configuration in both examples. No matter what version is used, they should be consistent.

bindings/python/tersets/__init__.py
```diff
@@ -136,7 +136,7 @@ def compress(values, method, error_bound: float) -> bytes:
     Examples:
       >>> import numpy as np
       >>> arr = np.random.random(10).astype(np.float64)
-      >>> blob = compress(arr, Method.SwingFilter, 0.01)
+      >>> blob = compress(arr, Method.SwingFilter, {"abs_error_bound": 0.01})
       >>> isinstance(blob, bytes)
```
skejserjensen: Unclear what the purpose of this line in the example is.

src/capi.zig
```diff
@@ -47,20 +45,23 @@ pub const Configuration = extern struct { method: u8, error_bound: f32 };
...
+    method_idx: u8,
+    configuration: [*:0]const u8,
 ) i32 {
     const uncompressed_values = uncompressed_values_array.data[0..uncompressed_values_array.len];
 
+    const configuration_slice: []const u8 = mem.span(configuration);
+
     // Returning 1 is equivalent to returning Error.UnknownMethod.
-    if (configuration.method > tersets.getMaxMethodIndex()) return 1;
+    if (method_idx > tersets.getMaxMethodIndex()) return 1;
```
skejserjensen: Why not write `index` instead of `idx` to make the code more clear? Again, I see no upsides to using shorthands instead of full words when naming variables, types, functions, etc. For me, the small amount of extra time needed to write and/or autocomplete and/or generate proper variables seems irrelevant compared to the benefit of making the code easier to read and thus easier to maintain.

README.md
```diff
@@ -35,11 +35,12 @@ pub fn main() void {
 
    // Configuration for compression.
    // The supported compression methods are specified in tersets.zig.
+   // The supported configuration are specified in config.zig.
    const method = tersets.Method.SwingFilter;
-   const error_bound: f32 = 0.1;
+   const configuration = "{ \"abs_error_bound\": 0.1 }";
```
skejserjensen: Can the use of JSON strings in Zig not be simplified by using [Zig multi-line strings](https://ziglang.org/documentation/master/#Multiline-String-Literals)? For example, something like [this](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:zig,selection:(endColumn:53,endLineNumber:4,positionColumn:53,positionLineNumber:4,selectionStartColumn:53,selectionStartLineNumber:4,startColumn:53,startLineNumber:4),source:'const+std+%3D+@import(%22std%22)%3B%0Apub+fn+main()+!!void+%7B%0A++++const+configuration_single+%3D+%22%7B+%5C%22abs_error_bound%5C%22:+0.1+%7D%22%3B%0A++++const+configuration_multi+%3D+%5C%5C%7B+%22abs_error_bound%22:+0.1+%7D%0A++++%3B%0A%0A++++std.debug.print(%22Single:+%7Bs%7D%5Cn+Mukti:+%7Bs%7D%5Cn%22,+.%7Bconfiguration_single,+configuration_multi%7D)%3B%0A%7D'),l:'5',n:'0',o:'Zig+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:z0151,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:zig,libs:!(),options:'',overrides:!((name:edition,value:'2024')),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+zig+0.15.1+(Editor+%231)',t:'0')),header:(),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'zig+0.15.1',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+zig+0.15.1+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4). This seems to be especially convinient if the compression methods takes multiple parameters.

src/config.zig
```diff
@@ -0,0 +1,272 @@
+// Copyright 2024 TerseTS Contributors
```
skejserjensen: Maybe rename the file to `configuration` to match what the parameter is called.

src/config.zig
```diff
@@ -0,0 +1,272 @@
...
+    AggregateError: AggregateError,
+    AreaUnderCurveError: AreaUnderCurveError,
+    InvalidConfiguration: InvalidConfiguration,
+    EmptyConfiguration: EmptyConfiguration,
+};
+
+/// Parses the `configuration` json string for the given compression `method`.
+/// Returns the appropriate configuration type or an `InvalidConfiguration` if parsing fails.
+/// The `allocator` is used for temporary allocations during parsing.
+pub fn parse(
```
skejserjensen: I agree that it makes sense that the code for parsing the configuration is done by a shared function so all compression methods can use it. But I suggest that the code for checking that the configuration contains everything needed should be part of each compression method so it is clear from the implementation of each compression method what parameters it contains and what values they should have. For example, this method currently checks that the keys are what each compression method needs, but the compression methods themselves check the value of the error bound. So I suggest performing the entire check in each compression method. Of course, a helper function that checks the configuration based on a list of required parameters should probably be implemented in this file and then be used by each of the compression methods instead of reimplementing such logic in each of the compression methods. Making this change should also allow the switch on `method` to be removed and the `Configuration` `union(enum)` to be removed as it would be clear from the parameters provided by each compression method what parameters are needed, and maybe it will even be possible to also perform value checks through a shared parameterized function. Thus, further reducing the amount of duplicate code in each compression method.

src/histogram_representation/constant_linear_representation.zig
```diff
@@ -59,17 +59,8 @@ pub fn compressPWCH(
     allocator: mem.Allocator,
     uncompressed_values: []const f64,
     compressed_values: *ArrayList(u8),
-    error_bound: f32,
+    max_buckets: u32,
```
skejserjensen: Change `max_buckets` to `maximum_buckets` in all files, there is no benefit to save characters.

src/histogram_representation/constant_linear_representation.zig
```diff
@@ -685,7 +671,7 @@ test "Random clusters, elements per cluster and values for PWCH" {
     const random = prng.random();
 
     const allocator = std.testing.allocator;
-    const num_cluster: usize = prng.random().uintLessThan(usize, 100) + 10;
+    const num_cluster: u32 = prng.random().uintLessThan(u32, 100) + 10;
```
skejserjensen: `number_of_clusters`, there is no benefit to save characters.

src/tester.zig
```diff
@@ -260,15 +260,26 @@ pub fn testGeneratedCompression(
...
+        .{ranged_error_bound},
+    );
+    defer allocator.free(config_json);
 
     const compressed = try tersets.compress(
         allocator,
         uncompressed_values.items,
         method,
-        ranged_error_bound,
+        config_json,
```
skejserjensen: Change `config_json` to `configuration_json` in all files, I see no benefit to saving these few characters.

README.md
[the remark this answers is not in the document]
cmcuza: Done!

README.md
```diff
@@ -86,15 +87,14 @@ int main(void) {
     printf("Uncompressed data length: %zu\n", uncompressed_values.len);
 
     // Configuration for compression.
-    // The C API mirrors the `Method` enum from TerseTS.
-    struct Configuration config = { SwingFilter, 0.1f };
+     const char *cfg = "{\"abs_error_bound\": 0.01}";
```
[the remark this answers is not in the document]
cmcuza: Sure!

bindings/python/tersets/__init__.py
cmcuza: I removed it.

README.md
cmcuza: I tried this a couple of times, but it looked weird to me. It looks even weirder in the `Readme.md` since it is rendered as a comment. I think users can use it if they want.

cmcuza: I have now used it across all of TerseTS. Still, it looks weird in the Readme.md

src/configuration.zig
```diff
@@ -0,0 +1,308 @@
...
+        => AggregateError,
+        .VisvalingamWhyatt => AreaUnderCurveError,
+        .PiecewiseConstantHistogram,
+        .PiecewiseLinearHistogram,
+        => HistogramBinsNumber,
+        .RunLengthEncoding => EmptyConfiguration,
+    };
+}
+
+test "check configuration parsing for all methods" {
```
skejserjensen: Maybe move the tests for parsing each compression method configuration to each compression method. For example, one compression method may use `RelativeErrorBound` and support an error bound of 0--100, while another may use `RelativeErrorBound` and only support an error bound of 10--100. So that parse configuration and ensure that they only accept the supported configurations for each compression method anyway. This should also allow `getConfiguration()` to be deleted so we only dispatch on `method` once in `tersets.compress()` and once in `tersets.decompress()` which I personally think is a nice simple design to understand and thus maintain.

src/config.zig
skejserjensen: I spend a few days thinking about the pros and cons of one big shared parsing function like was initially implemented in this PR and having a small generic parsing function called by each of the compression method like the PR currently used. I was unable to come up with pros for the initial design as it seemed to require more advanced code than the current approach and made it harder to understand which configuration a compression method uses, so I much prefer the current design.

skejserjensen (changes requested):

bindings/c/tersets.h
```diff
@@ -38,17 +38,12 @@ struct CompressedValues {
...
-  float   error_bound;
-};
-
 // Compress uncompressed_values to compressed_values according to configuration.
 // Returns 0 on success, non-zero on error (e.g., 1 = unsupported method).
 int32_t compress(struct UncompressedValues uncompressed_values,
                  struct CompressedValues *compressed_values,
-                 struct Configuration configuration);
+                 uint8_t method, 
+                 const char * configuration);
```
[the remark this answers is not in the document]
cmcuza: Done!

src/histogram_representation/constant_linear_representation.zig
```diff
@@ -309,7 +328,7 @@ const Histogram = struct {
     // Memory allocator for the convex hull in the buckets.
     allocator: mem.Allocator,
     // Target number of buckets.
-    max_buckets: usize,
+    max_buckets: u32,
```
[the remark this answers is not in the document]
cmcuza: Done!

src/configuration.zig
cmcuza: This is a good idea and I tried to implemented, but as you will see in the code, most the tests are almost always repeated. This is because the logic of wheter a confiuration works or not does not dependent on a method, but in general on the configuration itself. Right now, most of hte lossy compressors use ABS_ERROR_BOUND, and testing that the configurations for this configuration is indepedent of the method, e.g., Swing and PMC test for ABS_ERROR_BOUNS is the same. I then created small helper functions to avoid repeating the same function, and only call these functions per method, even if it is the same call.

cmcuza: I think the current solution is not good, but leaving here for better understanding of the full complexity. I could change everything to call the functions directly, but again, where the configuration is parsed or not it does not depend on each function.

src/configuration.zig
```diff
@@ -0,0 +1,308 @@
...
+
+/// This is a small convenience wrapper around json.parseFromSlice that accepts a
+/// `ConfigurationType` and parses the the JSON text `configuration`. The function
+/// returns the parsed value on success, or `null` on failure. The `allocator` is
+/// used by the JSON parser.
+pub fn parse(
+    allocator: Allocator,
+    comptime ConfigurationType: type,
+    configuration: []const u8,
+) ?ConfigurationType {
```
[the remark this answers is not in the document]
cmcuza: Done. Much better this way.

skejserjensen (changes requested):
See this comment: https://github.com/cmcuza/TerseTS/pull/104#discussion_r2527722972

skejserjensen (approved):

PR #107, opened 2025-11-20
Add ZigZag and EliasGamma encoders

cmcuza:
### Summary

This PR adds the ZigZag and EliasGamma encoders in the `shared_function.zig` file. These encoders are used for modern compressors like SerfQT [1]. I am adding these encoders as they contain enough code and logic by themselves to require an independent  PR. Once merged, adding SerfQT is much easier. 

### Files Added/Modified

- I have only modified the `shared_function.zig` file.

### Testing

- I have added tests for both ZigZag and EliasGamma encoders.

[1] Li, Ruiyuan, et al. "Serf: Streaming Error-Bounded Floating-Point Compression." Proceedings of the ACM on Management of Data 3.3 (2025): 1-27.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,12 @@ pub fn main() void {
 
    // Configuration for compression.
    // The supported compression methods are specified in tersets.zig.
+   // The supported configuration are specified in config.zig.
    const method = tersets.Method.SwingFilter;
-   const error_bound: f32 = 0.1;
+   const configuration = "{ \"abs_error_bound\": 0.1 }";
 
    // Compress the data.
-   var compressed_values = try tersets.compress(uncompressed_values, allocator, method, error_bound);
+   var compressed_values = try tersets.compress(uncompressed_values, allocator, method, configuration);
    // The compressed values point to dynamically allocated data that should be deallocated.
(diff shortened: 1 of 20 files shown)


src/utilities/shared_functions.zig
```diff
@@ -182,3 +183,178 @@ pub fn isApproximatelyEqual(value_a: f64, value_b: f64) bool {
...
+            tester.max_test_value,
+            default_random,
+        );
+        const encoded = zigzagEncoder(original);
+        const decoded = zigzagDecoder(encoded);
+        try testing.expect(decoded == original);
+    }
+}
+
+test "eliasGammaEncoder can encode simple values correctly" {
```
skejserjensen: Since both `eliasGammaEncoder()` and `eliasGammaDecoder()` has been implemented, maybe only keep _"eliasGammaEncoder cannot encode zero value"_ and _"eliasGammaDecoder can encode and decode complex values correctly"_ as these two tests together should ensure that all `u64` values are encoded correctly and that zero is rejected. Of course, if it is important that 15, 7, 3, 1 are encoded exactly as the test checks, maybe also keep the other two tests.

PR #148, opened 2026-05-24
Implement Chimp-64, and Chimp-128

Romercor:
### Summary

- Closes #22 
- Based on [Liakos et al., "Chimp: Efficient Lossless Floating Point Compression for Time Series Databases", VLDB 2022](https://doi.org/10.14778/3551793.3551852)
- CHIMP64 is the base XOR-based codec with 3-bit leading-zero buckets and a 6-bit trailing-zero threshold. CHIMP128 extends it with a 128-value ring buffer indexed by a 14-bit LSB hash, enabling O(1) predictor lookups that outperform the immediately-previous-value baseline on periodic time series.

### Files Added/Modified

- New method file: src/lossless_compression/chimp64.zig
- New method file: src/lossless_compression/chimp128.zig
- src/tersets.zig — added Chimp64 and Chimp128 enum variants, compress/decompress dispatch, extract/rebuild stubs, test-exclusion list entries
- src/configuration.zig — added empty configuration support for Chimp64 and Chimp128
- src/capi.zig — added enum value assertions (Chimp64 = 17, Chimp128 = 18)
- src/tester.zig — added expectLosslessRoundTrip helper for bit-exact round-trip testing via direct codec calls (can be used by all lossless codecs)
- bindings/c/tersets.h, bindings/julia/TerseTS.jl, bindings/python/tersets/__init__.py, bindings/rust/src/lib.rs — added Chimp64 and Chimp128 to method enums

### Testing

- Unit tests added in chimp64.zig and chimp128.zig: empty input, single value, two values, repeated values (XOR-zero marker path), changing values (bucket transitions), special floating-point values (NaN payloads, ±∞, floatMax), edge floats (±0.0, subnormals, nextAfter pairs)
- CHIMP128 additionally: sequence longer than the 128-value ring buffer to exercise slot wraparound and predictor reuse
- Compression-ratio tests: 500 identical values compress below raw 8 × N byte size for both methods
- Configuration parsing tests: empty configuration accepted, invalid (extra-field) configuration rejected with InvalidConfiguration
- Generated round-trip tests via the TerseTS dispatcher across all 14 data distributions, including NaN/infinity injection variants, verifying bitwise lossless recovery

PR #149, opened 2026-05-27
Fix/chimp review 2

Romercor:
### Summary

Addresses the second round of Copilot review comments on upstream PR #148:

- Reject meaningful_bit_count == 0 in the marker 01 decoder explicitly (both codecs) - encoder never emits this, so it can only come from a corrupted stream.
- Add a non-canonical NaN (0x7ff8000000000001) to the special-values test so it actually exercises the payload-preservation claim in the comment.
- Cast the LSB hash key to usize explicitly in chimp128 to drop the implicit u64 - usize narrowing.

### Files Added/Modified

- src/lossless_compression/chimp64.zig
- src/lossless_compression/chimp128.zig

### Testing

- zig build test --summary all → 182/182 pass.


Romercor:
wrong target, wanted to pull request to my fork first.

PR #151, opened 2026-06-05
Adding zig 0.17 compilation fixes

xiamaz:
### Summary

This PR fixes minor compilation issues that block compiling tersets using 0.17 (dev master) version of zig.

getLast change breaks 0.16 compatibility.

### Files Added/Modified

- sim piece - getLast now returns optional that needs to be unwrapped
- shared structs - use @Int instead of relying on meta.Int which has been removed 

### Testing

- nothing applies

### DIFF
--- a/src/lossy_compression/functional_approximation/sim_piece.zig
+++ b/src/lossy_compression/functional_approximation/sim_piece.zig
@@ -167,14 +167,15 @@ pub fn decompress(
         current_index = next_metadata_start_index;
     }
 
-    const current_metadata = segments_metadata.getLast();
-    try decompressSegment(
-        allocator,
-        current_metadata,
-        current_index,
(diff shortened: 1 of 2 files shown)


cmcuza:
Hi @xiamaz,

Thanks for the PR. It is almost ready, but you also need to update the workflows in GitHub to 0.17 at .github/workflows/zig-fmt-build-and-test-on-pr-and-push.yml. If you are in doubt, take a look at #146.

cmcuza:
Hi @xiamaz,
I just realized that version 0.17.0 is still under development. So, we'll stick with 0.16 until the Zig team releases the stable version. Thus, I will close the PR for now and reopen it once the stable 0.17 version is out. Thanks a lot.

PR #156, opened 2026-06-16
Improve configuration file and local workflow script

cmcuza:
### Summary

This PR replaces the logic of `getMaxMethodIndex` to run in constant time. The PR also improves the script to run the GitHub workflow locally before commiting.

### DIFF
--- a/src/tersets.zig
+++ b/src/tersets.zig
@@ -744,9 +744,7 @@ pub fn rebuild(
 
 /// Get the maximum index of the available methods in TerseTS.
 pub fn getMaxMethodIndex() usize {
-    const max_method_index: usize = @typeInfo(Method).@"enum".fields.len - 1;
-
-    return max_method_index;
+    return @typeInfo(Method).@"enum".fields.len - 1;
 }
 
 test "extract and rebuild works for any compression method supported" {
--- a/src/tersets.zig
+++ b/src/tersets.zig
@@ -744,14 +744,9 @@ pub fn rebuild(
 
 /// Get the maximum index of the available methods in TerseTS.
 pub fn getMaxMethodIndex() usize {
-    const method_info = @typeInfo(Method).@"enum";
+    const max_method_index: usize = @typeInfo(Method).@"enum".fields.len - 1;
 
-    var max_index: usize = 0;
-    for (method_info.fields, 0..) |_, i| {
-        max_index = if (i > max_index) i else max_index;
-    }
-
-    return max_index;
+    return max_method_index;
 }
 
 test "extract and rebuild works for any compression method supported" {
--- a/tools/run-github-workflow-local.sh
+++ b/tools/run-github-workflow-local.sh
@@ -14,7 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Mirrors .github/workflows/ci.yml for local verification before pushing.
+# Approximates .github/workflows/ci.yml for local verification before pushing.
+# The C-binding only runs CLang for testing C-syntax. Therefore, it is skipped in this
+# script to avoid requiring CLang to be installed locally.
+
 
 set -eu
 
@@ -30,8 +33,7 @@ print_step() {
 
 # Accumulates names of missing commands and reports them all at once.
 assert_command_available() {
-    local missing_commands=""
-    local command_name
+    missing_commands=""
 
     for command_name in "$@"
     do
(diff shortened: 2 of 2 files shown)


src/tersets.zig
```diff
@@ -744,14 +744,9 @@ pub fn rebuild(
 
 /// Get the maximum index of the available methods in TerseTS.
 pub fn getMaxMethodIndex() usize {
-    const method_info = @typeInfo(Method).@"enum";
+    const max_method_index: usize = @typeInfo(Method).@"enum".fields.len - 1;
```
skejserjensen: Maybe replace `const max_method_index: usize =` with `return` if possible.

tools/run-github-workflow-local.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
skejserjensen, line 1: Good to write the script in `POSIX shell` to be compatible with more shells than just `bash`, but then the script should also stick strictly to `POSIX shell` and not use any features from `bash`. For example,  `POSIX shell` `local` is undefined. Also, be sure to test the script on a system where `sh` is actually a `POSIX shell`, unfortunately, many operating systems make `sh` run a more featureful shell like `bash` or `zsh`. For example, on macOS, it is `bash` and `sh -version` prints `GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)`. FreeBSD uses the Bourne shell-compatible `sh` as the default user shell, so it is fairly simple to test the script using FreeBSD. For more information, see, e.g., these resources:

- https://www.shellcheck.net/
- https://mywiki.wooledge.org/Bashism
- https://linux.die.net/man/1/checkbashisms

tools/run-github-workflow-local.sh
```diff
@@ -54,7 +52,7 @@ assert_command_available zig gcc julia python rustup cargo
 
 cd "$repo_root"
 
-# Build the native library first so local bindings can load zig-out artifacts.
+# Build in both modes first so binding tests can load the zig-out artifacts.
```
skejserjensen: Does Zig actually have both artifacts available after these steps? Does the second `zig build` not overwrite the artifacts, as the build mode is not part of the name? If so, maybe split the comment into two lines, one that specifies why both build modes are used and one that specifies what artifacts are available and for whom.

tools/run-github-workflow-local.sh
```diff
@@ -66,17 +64,14 @@ zig build test -Doptimize="Debug"
 print_step "Zig Build Test ReleaseFast"
 zig build test -Doptimize="ReleaseFast"
 
-# Julia
```
skejserjensen, line 69: Why remove all the headers that make it clear what language each part is for? I thought it made the script more readable to have a label above the section for each language.

skejserjensen (changes requested):

src/tersets.zig
cmcuza: Done.

tools/run-github-workflow-local.sh
cmcuza: True, you are right. Fixed.

tools/run-github-workflow-local.sh
cmcuza: Since we were printing, I thought it was redundant. I brought it back.

tools/run-github-workflow-local.sh
cmcuza: This was useful! Fixed.

tools/run-github-workflow-local.sh
```diff
@@ -31,21 +28,22 @@ print_step() {
     printf "\n==> %s\n" "$1"
 }
 
+# Accumulates names of missing commands and reports them all at once.
 assert_command_available() {
-    local missing_commands=()
+    local missing_commands=""
     local command_name
```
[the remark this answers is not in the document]
cmcuza: Fixed.

tools/run-github-workflow-local.sh
```diff
@@ -14,14 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This script is intended to be run from the root of the repository 
-# and will execute the same steps as the GitHub workflow defined in
-# .github/workflows/ci.yml, allowing you to verify that the workflow will
-# succeed locally before pushing changes.
+# Mirrors .github/workflows/ci.yml for local verification before pushing.
```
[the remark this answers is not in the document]
cmcuza: Fixed.

tools/run-github-workflow-local.sh
```diff
@@ -14,14 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This script is intended to be run from the root of the repository 
-# and will execute the same steps as the GitHub workflow defined in
-# .github/workflows/ci.yml, allowing you to verify that the workflow will
-# succeed locally before pushing changes.
+# Approximates .github/workflows/ci.yml for local verification before pushing.
+# The C-binding only runs CLang for testing C-syntax. Therefore, it is skipped in this
```
skejserjensen: [Clang](https://en.wikipedia.org/wiki/Clang) is usually written with lowercase l.

tools/run-github-workflow-local.sh
```diff
@@ -14,14 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This script is intended to be run from the root of the repository 
-# and will execute the same steps as the GitHub workflow defined in
-# .github/workflows/ci.yml, allowing you to verify that the workflow will
-# succeed locally before pushing changes.
+# Approximates .github/workflows/ci.yml for local verification before pushing.
```
skejserjensen, line 17: Maybe check if the user has a C compiler installed and then run, `cl.exe`, `gcc`, `clang`, `icc`, etc so this script match GitHub Actions directly.

skejserjensen (changes requested):

tools/run-github-workflow-local.sh
cmcuza: fixed.

tools/run-github-workflow-local.sh
cmcuza: Fixed.

skejserjensen (approved):

